PR #68, opened 2026-04-16
Add direct deployment flow

Fermionic-Lyu:
### Add direct deployment flow with fallback to legacy zip-based upload
- Adds a direct file upload deployment path in [`deploy.ts`](https://github.com/InsForge/CLI/pull/68/files#diff-29891bf16de6dea65a788dd800bed2aea07c75a8c9de36a4d1abe448d6cf2ab2): scans the source directory, creates a deployment session via `POST /api/deployments/direct`, uploads only non-already-uploaded files concurrently (up to 8 at a time), then starts and polls the deployment.
- The orchestrator tries the direct path first and falls back to the existing zip-based flow when the backend returns 404, signaled by the new `DirectDeploymentUnsupportedError`.
- Adds file manifest helpers: `collectDeploymentFiles` (recursive directory scan with exclusions), `hashFile` (SHA-1 + byte size), and `runWithConcurrency` for bounded parallel uploads.
- Extracts polling into a standalone `pollDeployment` function that optionally triggers a sync endpoint before reading status, useful for non-cloud backends.
- The `--env` flag now accepts either an array of key/value pairs or a plain object, coercing values to strings.

<sup><a href="https://app.macroscope.com">Macroscope</a> summarized c183ba7.</sup>

## Summary by CodeRabbit

* **New Features**
  * Deployment now supports direct file uploads for improved efficiency
  * Environment variable flag accepts both array and object formats

* **Improvements**
  * Concurrent file uploading (up to 8 files simultaneously) for faster deployments
  * Automatic fallback to legacy upload method when direct upload unavailable
  * Enhanced error handling and status synchronization during deployment

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,16 +1,16 @@
 {
   "name": "@insforge/cli",
-  "version": "0.1.47",
+  "version": "0.1.51-dev.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@insforge/cli",
-      "version": "0.1.47",
+      "version": "0.1.51-dev.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@clack/prompts": "^0.9.1",
-        "@insforge/shared-schemas": "^1.1.44",
+        "@insforge/shared-schemas": "^1.1.49",
         "archiver": "^7.0.1",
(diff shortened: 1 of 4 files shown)


tonychang04 (approved):
sanity check everything looks fine!

PR #69, opened 2026-04-17
fix(analytics): flush PostHog before process.exit in error paths

jwfing:
## Summary
- `handleError()` calls `process.exit()` synchronously, terminating the process before the `finally { await shutdownAnalytics() }` block gets a chance to flush PostHog's batched events. On any error path the event was silently dropped.
- This is why `cli_diagnose_invoked` with `subcommand='ai'` never showed up in PostHog — the AI streaming path tends to hit the error handler (e.g. backend returns an error SSE event, or non-TTY interactive rating prompt gets killed), while the other subcommands usually complete successfully and their finally block flushes normally.
- Fix: add `await shutdownAnalytics()` in the catch block *before* `handleError` runs. Applied to every action handler that emits a PostHog event and can reach `handleError` — the 5 diagnose commands, `create`, and `link` (including `link`'s nested OSS direct-connect catch, which previously skipped the outer finally too).
- Bumps version to 0.1.52 for release.

Verified locally: after rebuilding and running `diagnose --ai`, the `subcommand='ai'` event now appears in PostHog.

## Test plan
- [x] `npm run lint` passes (already verified locally)
- [x] Build with `POSTHOG_API_KEY=<real_key> npm run build`
- [x] Run `node dist/index.js diagnose --ai "test question"` against a linked project that triggers the streaming error path, and confirm the `cli_diagnose_invoked` event with `subcommand='ai'` appears in PostHog
- [x] Confirm success-path events (`subcommand='report'`, `db`, `logs`, `advisor`, `metrics`) still appear as before

## Follow-up
The root cause is structural — any new command that follows the same `catch { handleError } finally { shutdownAnalytics }` pattern will silently drop events on error. Consider a follow-up refactor that makes `handleError` async and awaits flushing internally, so call sites can't forget. Left out of this PR to keep the fix surgical.

> [!NOTE]
> ### Flush PostHog analytics before `process.exit` in error paths across CLI commands
> Adds `await shutdownAnalytics()` in the `catch` blocks of `create`, `diagnose`, `diagnose advisor`, `diagnose db`, `diagnose logs`, `diagnose metrics`, and `projects link` commands, immediately before `handleError()` is called. Previously, `shutdownAnalytics()` only ran in `finally`, but `handleError` can call `process.exit` before the `finally` block completes, causing analytics events to be dropped on error paths.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized a817281.</sup>
> 
>

## Summary by CodeRabbit

* **Chores**
  * Version bumped to 0.1.52
  * Improved error handling consistency and resource cleanup procedures across command handlers
  * Enhanced shutdown reliability during unexpected failures to maintain application stability
  * Optimized error recovery processes to ensure proper cleanup in all execution scenarios

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@insforge/cli",
-  "version": "0.1.51",
+  "version": "0.1.52",
   "description": "InsForge CLI - Command line tool for InsForge platform",
   "type": "module",
   "bin": {
--- a/src/commands/create.ts
+++ b/src/commands/create.ts
@@ -484,6 +484,7 @@ export function registerCreateCommand(program: Command): void {
           throw err;
         }
       } catch (err) {
+        await shutdownAnalytics();
         handleError(err, json);
       } finally {
         await shutdownAnalytics();
--- a/src/commands/diagnose/advisor.ts
+++ b/src/commands/diagnose/advisor.ts
@@ -112,6 +112,7 @@ export function registerDiagnoseAdvisorCommand(diagnoseCmd: Command): void {
         await reportCliUsage('cli.diagnose.advisor', true);
       } catch (err) {
         await reportCliUsage('cli.diagnose.advisor', false);
+        await shutdownAnalytics();
         handleError(err, json);
       } finally {
         await shutdownAnalytics();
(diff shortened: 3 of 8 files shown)


tonychang04 (approved):
LGTM. Clean fix — `shutdownAnalytics()` before `handleError()` in every catch block that can reach `process.exit`. Applied consistently across all 8 files. The double-flush (catch + finally) is harmless since `shutdownAnalytics` is idempotent.

Agree with the follow-up note about making `handleError` async with internal flushing to prevent future call sites from hitting this same issue.

PR #77, opened 2026-04-21
feat(link): support --template in direct-link mode

tonychang04:
## Summary

`link --template <name>` was advertised in `--help` but silently ignored when combined with `--api-base-url` + `--api-key`. The direct-link branch did `saveProjectConfig` → `installSkills` → `return` before ever touching the template handling code. This meant the agent-facing signup + link flow (used by the public skill.md on insforge.dev) had to fall back to a six-step manual recipe — clone template, `cd` in, link, fetch anon key via HTTP, write `.env.local`, run migration, npm install — to do what the OAuth path already does in one command.

This PR wires the direct-link branch into the same template flow used by the OAuth path.

## Changes in `src/commands/projects/link.ts`

- Pull template-name validation to the top of the action (shared between both paths). Invalid names fail before any I/O.
- Hoist `githubTemplates` to action scope so both paths reference the same list.
- When `--template` is set alongside `--api-base-url` + `--api-key`:
  - Prompt for a directory name (default `insforge-<template>` since direct-link has no `project.name` from the platform API).
  - `mkdir` / `chdir` into the new directory.
  - `saveProjectConfig` inside the template dir (so the CLI's CWD-only `.insforge/project.json` lookup finds it).
  - Call `downloadGitHubTemplate` (for `chatbot` / `crm` / `e-commerce` / `todo` / `react-github-copy` / `nextjs-github-copy`) or `downloadTemplate` (for the `react` / `nextjs` `create-insforge-app` scaffolds).
  - `npm install` in the new directory.
  - `installSkills`, record the existing `cli.link_direct` telemetry event (with `template` added to the payload), emit the same "What's next" note as the OAuth template flow.
- Non-template direct-link behavior is preserved exactly as-is.
- OAuth path is untouched except for two tiny refactors: it now uses the hoisted `validTemplates` and `githubTemplates` instead of re-declaring them.

## Why the default dir is `insforge-<template>`

OAuth defaults to `project.name` because `listProjects` returned it. Direct-link uses sentinel values (`FAKE_PROJECT_ID`, `project_name: 'oss-project'`, `appkey: 'ossfkey'`), so that's a poor default. `insforge-<template>` matches the convention `create-insforge-app` uses for its own default directory when `--projectName` is omitted.

## Verification

- [x] `npx tsc --noEmit` — clean on `link.ts` (pre-existing errors elsewhere in the repo are unrelated).
- [x] `npm run lint` — vitest 92/92 tests pass, eslint 0 errors.
- [x] `tsup` build succeeds.
- [x] Built `dist/index.js` contains the new branch at lines 2160–2241 with the correct control flow.
- [x] `link --help` output unchanged.
- [ ] **End-to-end smoke test against a real project** — not run from my sandbox. Please verify before publish:
  ```bash
  npx @insforge/cli link --api-base-url <url> --api-key <key> --template chatbot
  ```
  Expected: prompts for dir name (default `insforge-chatbot`), clones the chatbot template into that dir, writes `.env.local` with real `projectUrl` + anon key + `NEXT_PUBLIC_APP_URL`, applies `migrations/db_init.sql`, runs `npm install`, prints the "What's next" note.

## Follow-up

Once this is published, the public `skill.md` flow (insforge-cloud PR #376) can be simplified from the six-step manual recipe to a single `link --template` call.

## Summary by CodeRabbit

* **New Features**
  * Validates template selection during project linking to prevent invalid choices.
  * Direct API linking now supports creating a template-based project directory, downloading the template, saving configuration, and running initial setup (including dependency install when applicable).
  * OAuth linking retains guided template setup and post-configuration "What's next" guidance.

* **Chores**
  * Package version updated to 0.1.55.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@insforge/cli",
-  "version": "0.1.53",
+  "version": "0.1.55",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@insforge/cli",
-      "version": "0.1.53",
+      "version": "0.1.55",
       "license": "Apache-2.0",
       "dependencies": {
         "@clack/prompts": "^0.9.1",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@insforge/cli",
-  "version": "0.1.54",
+  "version": "0.1.55",
   "description": "InsForge CLI - Command line tool for InsForge platform",
   "type": "module",
   "bin": {
--- a/src/commands/projects/link.ts
+++ b/src/commands/projects/link.ts
@@ -39,6 +39,14 @@ export function registerProjectLinkCommand(program: Command): void {
     .option('--api-key <key>', 'API Key for direct linking (OSS/Self-hosted)')
     .action(async (opts, cmd) => {
       const { json, apiUrl } = getRootOpts(cmd);
+
+      // Shared template validation — applies to both direct-link and OAuth paths.
+      const validTemplates = ['react', 'nextjs', 'chatbot', 'crm', 'e-commerce', 'todo'];
+      const githubTemplates = ['chatbot', 'crm', 'e-commerce', 'nextjs', 'react', 'todo'];
+      if (opts.template && !validTemplates.includes(opts.template)) {
+        throw new CLIError(`Invalid template "${opts.template}". Valid options: ${validTemplates.join(', ')}`);
+      }
+
       try {
         if (opts.apiBaseUrl || opts.apiKey) {
           try {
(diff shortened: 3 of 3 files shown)


tonychang04:
Review cleanup pushed in 834e412:

1. **Removed the duplicate `githubTemplates` list.** As @reviewer correctly noted, `validTemplates` and `githubTemplates` held the same six names here. The split only made sense in `create.ts` where `validTemplates` also includes `'empty'` (meaning "no template"). With the split collapsed, the `downloadTemplate` else-branch in both link paths becomes dead code — every validated template value maps to a GitHub-hosted template — so it's deleted. Also removed the now-unused `downloadTemplate` / `Framework` imports.
2. **Fixed a validation scoping bug I introduced.** The hoisted `--template` validation was sitting *outside* the outer try/catch, so an invalid template name would have bubbled up as an uncaught promise rejection instead of being routed through `handleError` like every other failure. Moved it inside the outer try.

No behavior change for valid template names or for the non-template paths. All checks still green: typecheck, eslint, 92/92 vitest, tsup build.

Fermionic-Lyu (approved):

PR #87, opened 2026-04-26
feat(cli/compute): compute deploy [dir] — source mode (local docker build + push)

tonychang04:
…ired)

Replaces the v1 flyctl-based compute deploy with the v3 source-upload flow (Phase 4 of compute v3):

  insforge compute deploy ./my-app --name my-api

Detects Dockerfile in [dir]; tars the directory (excluding .git, node_modules, etc.); uploads source.tgz directly to InsForge's S3 staging bucket via a presigned PUT URL minted by cloud; cloud triggers AWS CodeBuild to build the image; deploys the resulting image to Fly.

What this means for the user:
- No flyctl install required (was previously mandatory)
- No FLY_API_TOKEN required (cloud holds it)
- No local Docker required (CodeBuild builds it)
- Just write a Dockerfile and run one command

What changes for image-mode users:
- Nothing. compute deploy --image <url> --name x works identically.

Architecture:
- Bytes go laptop -> S3 directly via presigned URL (never proxy through OSS or cloud)
- CLI calls only KB-sized JSON (build-creds + create-service requests)

New: src/lib/upload.ts (tar + curl PUT helpers), 3 unit tests. Bumps version 0.1.58 -> 0.1.59.

## Summary by CodeRabbit

* **New Features**
  * Added ability to deploy from a local source (builds and pushes a container) or from a pre-built image; deploy prints endpoint URL when available

* **Refactor**
  * Deploy split into two exclusive modes (image vs source) with stricter validation and sensible defaults for port/cpu/memory/region; `--env` must be JSON

* **Removed**
  * Removed the separate "create" command; workflows now use "deploy"

* **Tests**
  * Added tests for Docker build/login/push behaviors and error handling

* **Chores**
  * Version bumped to 0.1.59


jwfing (changes requested):
## Summary

The PR restructures `compute deploy` into image-mode and source-mode, and introduces `src/lib/docker.ts` for local Docker build+push. Two blocking issues prevent merging as-is.

---

## Requirements context

No spec or plan file under `docs/specs/` matches this PR's scope (the directory contains only `diagnose` and `db-migrations` specs). Assessment is based on the PR description and code diff alone.

---

## Findings

### Critical

#### Software Engineering — Missing unit tests
`src/lib/docker.ts` is a new module with four exported functions (`ensureDockerAvailable`, `dockerBuild`, `dockerLogin`, `dockerPush`). There is no test file for it — no `docker.test.ts` anywhere in the repo. The PR description claims _"3 unit tests"_ exist, but they are absent from the diff.

Per project TDD conventions, every new behavior must have at least one test. The `dockerLogin` function in particular has non-trivial branching (success vs. `r.error`, `r.status !== 0`) and handles a short-lived auth token — exactly the kind of code that benefits from unit tests with a mocked `spawnSync`.

> **Blocking until tests are added for `src/lib/docker.ts`.**

#### Functionality — PR description directly contradicts the implementation

The PR body says:

> "No local Docker required (CodeBuild builds it)"  
> "tars the directory … uploads source.tgz directly to InsForge's S3 staging bucket via a presigned PUT URL … cloud triggers AWS CodeBuild to build the image"  
> "New: src/lib/upload.ts (tar + curl PUT helpers)"

The actual code does the opposite: it runs **`docker build` + `docker push` locally**, and the inline comment confirms it: _"Requires Docker locally."_ There is no `src/lib/upload.ts`, no tar, no S3 upload, no CodeBuild path.

This isn't a minor description lag. The stated user benefit _"No local Docker required"_ is outright wrong for source mode, and the architecture description (S3 presigned URL → CodeBuild) describes a completely different system than what ships. Anyone reading this PR — including users reading release notes generated from it — will be misled.

> **Either update the description/architecture to match the code, or confirm the wrong branch was pushed.**

---

### Suggestion

#### Software Engineering — `dockerBuild` / `dockerPush` skip `r.error` check (`src/lib/docker.ts:35, 55`)

`ensureDockerAvailable` correctly tests `r.error || r.status !== 0`. The downstream build and push helpers only check `r.status !== 0`:

```ts
// docker.ts:35
if (r.status !== 0) {
```

If `spawnSync` itself fails to fork (e.g., `ENOMEM` or a transient OS hiccup after the availability check passes), `r.error` is set and `r.status` is `null`. `null !== 0` happens to be `true`, so the error is still thrown — but the message will say `"exit null"` rather than the underlying OS error. Add `r.error || r.status !== 0` for both functions, and include `r.error?.message` in the thrown message.

#### Software Engineering — `void _omit` pattern in image mode (`src/commands/compute/deploy.ts:~107`)

```ts
const { name: _omit, ...updateBody } = body;
void _omit;
```

This is an unusual construct. Prefer a typed spread that makes the intent explicit:

```ts
const { name: _name, ...updateBody } = body as typeof baseBody & { name: string };
```

Or, since `body` already is `Record<string, unknown>`, simply:

```ts
// eslint-disable-next-line @typescript-eslint/no-unused-vars
const { name: _name, ...updateBody } = body;
```

Either way, a short comment explaining _why_ `name` is stripped from the PATCH body would help future readers.

#### Software Engineering — `package-lock.json` version gap

`package.json` bumps `0.1.58 → 0.1.59`, but `package-lock.json` jumps `0.1.57 → 0.1.59`, skipping `0.1.58`. This is almost certainly a rebase artifact (the lockfile was last committed when `package.json` was at `0.1.57`). Run `npm install` on the branch to regenerate the lockfile so both files agree throughout their history.

---

### Information

#### Performance — Local image not cleaned up after push (`src/commands/compute/deploy.ts`)

`dockerPush` leaves the built image (`registry.fly.io/<flyAppId>:cli-<timestamp>`) in the user's local Docker daemon. Over repeated deploys this accumulates. Consider logging a tip like `"Tip: run docker rmi <imageRef> to reclaim disk space"` after a successful push, or documenting this in the help text.

#### Security — `dockerLogin` error output (`src/lib/docker.ts:47-50`)

```ts
`docker login ${registry} failed (exit ${r.status}): ${(r.stderr || r.stdout || '').trim().slice(0, 300)}`
```

The deploy token is passed to Docker via `stdin` (good — not a CLI arg), and Docker does not echo `--password-stdin` input back in its error messages in any known version. The risk is low, but the pattern of forwarding raw Docker output into a `CLIError` (which may be forwarded to structured logs by `handleError`) is worth keeping in mind if the error format changes.

No security-relevant issues beyond the note above: all API paths use `encodeURIComponent`, user input to `--env` is validated as a JSON object, `spawnSync` is called with array arguments (no shell injection), and the deploy token is scoped to 20 minutes.

---

## Verdict

`request_changes` — two Critical items must be resolved:
1. Add unit tests for `src/lib/docker.ts`.
2. Reconcile (or rewrite) the PR description so it accurately reflects that source mode **requires local Docker** and uses a build+push flow, not S3+CodeBuild.

tonychang04:
Round-2 review fixes pushed in `9d4b58e`.

### Addressed

- **CodeRabbit (Major):** `docker.test.ts` — hoisted `spawnSyncMock` via `vi.hoisted()`. Per Vitest 4.x docs the `vi.mock` factory can't reference a top-level `const` (TDZ risk). All 110 tests still green after the change.
- **CodeRabbit (Minor):** `deploy.ts` — replaced `${absDir}/Dockerfile` template literal with `join(absDir, 'Dockerfile')` for Windows compat.
- **CodeRabbit (Minor):** `deploy.ts` — added per-key string-type validation on `--env` values; surfaces `--env values must be strings — got <type> for key "<k>"` at the input boundary instead of letting the API reject it.
- **CodeRabbit (Major):** `compute.test.ts` — added a 60s bounded poll for `running` status after the deploy assertion so the downstream `compute get` test doesn't race the async Fly boot.
- **jwfing (Info):** `deploy.ts` — print `Tip: docker rmi <imageRef> to reclaim local disk space` after successful source-mode deploy.

### Skipped (with reason)

- **CodeRabbit (Critical) "wrong flow / advertised feature not delivered":** This was based on the original PR description that mistakenly described an abandoned Path B (S3 + CodeBuild). Description rewritten in the previous round; the implementation (local docker build + push to `registry.fly.io` with per-app deploy token) IS the intended Path A.
- **CodeRabbit on `src/lib/upload.ts` (.env excludes, 429 build-creds handling, presigned URL):** That file was deleted as part of the Path A switch (commit `24ed35e`). Comments fired against the old commit's diff.
- **CodeRabbit (Nitpick) "list-then-dispatch may not scale":** Accept as-is — at the 5-services-per-project quota a server-side `?name=` lookup is premature.
- **jwfing (Info) `dockerLogin` error output security note:** Token is piped via `--password-stdin` (never on argv); Docker doesn't echo stdin auth in error messages. Keeping current error formatting.

### Verification
- `npm run lint` → 110 passed / 13 skipped, eslint clean
- `npm run build` → tsup ESM + DTS green

Re-requesting review.

jwfing (commented):
## Code Review — `feat(cli/compute): compute deploy [dir]` (source mode)

**Summary:** Solid implementation of compute v3.1 source-deploy (Path A: local docker build + push to `registry.fly.io`). No critical findings — four suggestions worth addressing before the next release, plus two informational notes.

---

### Requirements context

No matching spec or plan found under `/docs/superpowers/` — this directory does not exist in the repo. Assessment is based on the PR title/body, inline code comments, and DEVELOPMENT.md conventions.

> **Note:** The PR _body_ describes a different architecture (S3 presigned upload → CodeBuild → Fly) from what was implemented (local `docker build + push` to `registry.fly.io`). The PR _title_ ("local docker build + push") and the in-code comment (`// Path A`) are accurate. The body appears to be a stale draft. See Information §1.

---

### Findings

#### Critical

*(none)*

---

#### Suggestion

**[Software engineering / functionality] `--json` mode pollutes stdout in source mode**
`src/lib/docker.ts:33` (`dockerBuild`) and `src/lib/docker.ts:60` (`dockerPush`) both use `stdio: 'inherit'`, which streams docker build/push output directly to the process's stdout regardless of the `--json` flag. DEVELOPMENT.md explicitly requires respecting `--json` (line 25: _"Respect the `--json` flag via `getRootOpts(cmd)`"_).

In `--json` mode the output looks like:
```
Step 1/5 : FROM node:18
...
Successfully tagged registry.fly.io/my-app:cli-1714000000000
The push refers to repository [registry.fly.io/my-app]
...
{"name":"my-api","status":"running",...}   ← intended JSON, now unparseable
```

Any CI script or tool using `compute deploy [dir] --json` will receive malformed stdout. `dockerLogin` correctly uses `stdio: ['pipe','pipe','pipe']` — the same approach should be available for build/push when silent mode is needed. A `silent?: boolean` opt on `BuildOptions` (and a matching param on `dockerPush`) that switches to `stdio: 'pipe'` when truthy would fix this. Caller at `src/commands/compute/deploy.ts:197,203` would pass `silent: json`.

---

**[Software engineering] Version bump should be minor, not patch — breaking change**
`src/commands/compute/create.ts` is deleted, removing the `compute create` command. Any user or script running `insforge compute create` after updating will receive an unknown-command error. DEVELOPMENT.md (line 96–97) states: _"Follow semver — patch for fixes, minor for additive features, **major for breaking changes**."_ At the project's current `0.x` stage a minor bump (`0.1.58 → 0.2.0` or `0.1.58 → 0.1.59`+note) is the minimum expected. The PR bumps only a patch (`0.1.58 → 0.1.59`).

---

**[Software engineering] `agent-skills` repo not updated**
DEVELOPMENT.md lines 71–77 require updating `InsForge/agent-skills` whenever a command is added or significantly changed, so that AI agents relying on that skill don't give users wrong answers. This PR removes `compute create` entirely and overhauls `compute deploy` with a new `[dir]` argument and source mode — both changes that need to be reflected in `agent-skills`. No corresponding `agent-skills` PR is referenced.

---

**[Software engineering / functionality] No integration test for source mode**
`src/integration/compute.test.ts` tests image mode only. The new source-deploy path (service provision → deploy-token mint → `dockerBuild` → `dockerLogin` → `dockerPush` → PATCH) has no integration-level coverage. Unit tests in `src/lib/docker.test.ts` are thorough and well-structured; an integration test for the full source flow (even behind a `DOCKER_AVAILABLE` guard similar to `integrationEnabled`) would give confidence that the API contract between CLI and cloud is correct end-to-end.

---

#### Information

**[Functionality] PR body describes wrong architecture**
The description reads: _"tars the directory… uploads source.tgz directly to InsForge's S3 staging bucket via a presigned PUT URL… cloud triggers AWS CodeBuild to build the image."_ The actual implementation does local `docker build + push` to `registry.fly.io` (Path A). The body also references `src/lib/upload.ts` (doesn't exist) — the actual new file is `src/lib/docker.ts`. Recommend updating the body to match what was shipped to avoid confusion for future readers.

---

**[Software engineering] `package-lock.json` was already out of sync on `main`**
On `main`, `package.json` was at `0.1.58` but `package-lock.json` still showed `0.1.57`. This PR brings them back in sync at `0.1.59` as a side effect, but the pre-existing skew suggests a prior version bump was committed without re-running `npm install`. Worth noting for the team's release hygiene.

---

### Verdict

**approved** *(informational — no Critical findings; four Suggestions above are worth addressing before the next patch/minor release, in particular the `--json` stdout pollution and the `agent-skills` sync)*

jwfing (approved):
LGTM, approved.

PR #89, opened 2026-04-27
feat(cli/compute): source mode uses flyctl remote builder (no local Docker)

tonychang04:
## Summary

Source-mode deploy (`compute deploy <dir>`) no longer needs Docker installed locally. The CLI now shells out to `flyctl deploy --remote-only --build-only` using the same per-app deploy token the cloud already mints. Build runs on Fly's remote builder; the resulting image is pushed straight to `registry.fly.io`. Cloud still owns machine launch.

User-facing requirement drops from **Docker Desktop (~1 GB) + the InsForge CLI** to **single `flyctl` binary on PATH (~30 MB) + the InsForge CLI**.

Image mode (`compute deploy --image <url>`) is unchanged.

## Companion PRs

- **Cloud** (no-Docker live e2e + Round-3 fixes): https://github.com/InsForge/insforge-cloud-backend/pull/456
- **OSS provider** (comment update): https://github.com/InsForge/InsForge/pull/1156
- **Skills doc** (flyctl pattern): https://github.com/InsForge/insforge-skills/pull/45

## Why this is safe

The cloud-side `createDeployToken` already attenuates the macaroon with `IfPresent { ifs: [Apps[<oneApp>], FeatureSet[builder, wg]], else: deny }` + 20-min ValidityWindow. The token can drive the Fly remote builder (`FeatureSet[builder]`) and push to `registry.fly.io/<app>` (`Apps[<oneApp>]`), but cannot reach any other app or any org-level read endpoint. Verified end-to-end by 14 assertions in the cloud PR's live e2e.

## Test plan

- [x] vitest passes for compute + lib suites
- [x] `tsc --noEmit -p .` — no new type errors from this PR
- [ ] Reviewer: smoke-test `compute deploy ./some-dir --name foo` against staging cloud once #456 is merged

## Summary by CodeRabbit

* **Changes**
  * Deployment now uses Flyctl remote build-and-push instead of local Docker builds.
  * Flyctl CLI is required for source-mode deployments; Docker is no longer needed locally.
  * Progress and help messaging updated for the new workflow; no local image cleanup required after deployment.

### DIFF
--- a/src/commands/compute/deploy.ts
+++ b/src/commands/compute/deploy.ts
@@ -198,7 +198,7 @@ export function registerComputeDeployCommand(computeCmd: Command): void {
         //    attenuated FLY_API_TOKEN we just received.
         const imageLabel = `cli-${Date.now()}`;
         if (!json) outputInfo(`Building & pushing on Fly remote builder...`);
-        const { imageRef } = flyctlBuildAndPush({
+        const { imageRef } = await flyctlBuildAndPush({
           dir: absDir,
           appId: flyAppId,
           imageLabel,
--- a/src/commands/compute/deploy.ts
+++ b/src/commands/compute/deploy.ts
@@ -7,29 +7,31 @@ import { handleError, getRootOpts, CLIError } from '../../lib/errors.js';
 import { outputJson, outputSuccess, outputInfo } from '../../lib/output.js';
 import { reportCliUsage } from '../../lib/skills.js';
 import {
-  ensureDockerAvailable,
-  dockerBuild,
-  dockerLogin,
-  dockerPush,
-} from '../../lib/docker.js';
+  ensureFlyctlAvailable,
+  flyctlBuildAndPush,
+} from '../../lib/flyctl.js';
 
 // `compute deploy` has two modes:
 //
 //   1. Image mode (--image <url>):
-//      Deploy a pre-built image from any registry. Same as v1.
+//      Deploy a pre-built image from any registry. No flyctl/Docker needed.
 //
-//   2. Source mode ([dir]) — Path A (compute v3.1):
-//      CLI runs docker build + push to registry.fly.io using a per-app
-//      deploy token minted by the cloud. Cloud then launches the machine
-//      pointing at the freshly-pushed image. Requires Docker locally.
-//      Image bytes never proxy through OSS or cloud.
+//   2. Source mode ([dir]) — Path A (compute v3.2):
+//      CLI runs `flyctl deploy --remote-only --build-only` against the user's
+//      source directory using a per-app, attenuated deploy token minted by
+//      the cloud. The remote builder runs on Fly's infrastructure and pushes
+//      the image straight to registry.fly.io/<app>:<tag>. The cloud then
+//      launches the machine pointing at the freshly-pushed image.
+//      Requires `flyctl` on PATH (no Docker daemon needed).
+//      The deploy token is scoped to one app + builder/wg, with `else: deny`
+//      — it cannot deploy or read anything else in the InsForge Fly org.
 export function registerComputeDeployCommand(computeCmd: Command): void {
   computeCmd
     .command('deploy [dir]')
     .description(
       'Deploy a compute service. Two modes:\n' +
-        '  compute deploy <dir> --name <name>             (source mode — local docker build + push, requires Docker)\n' +
-        '  compute deploy --image <url> --name <name>     (image mode — deploys pre-built image, no Docker required)'
+        '  compute deploy <dir> --name <name>             (source mode — flyctl remote build + push, requires flyctl on PATH; no Docker needed)\n' +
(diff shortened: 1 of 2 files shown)


jwfing (changes requested):
## Summary

The flyctl-remote-builder approach is architecturally sound and the token-scoping story is convincing, but `src/lib/flyctl.ts` ships with zero unit tests — a clear violation of the project's TDD conventions that mirrors the exact parallel that already exists for `docker.ts`.

---

## Requirements context

No spec/plan file under `/docs/specs/` (or `/docs/superpowers/`) matches the scope of this PR. The three spec files present cover DB migrations and the `diagnose` command. Assessment is therefore based on the PR description alone.

---

## Findings

### Critical

#### [Software Engineering] Missing unit tests for `src/lib/flyctl.ts`

`src/lib/flyctl.ts` introduces two exported functions — `ensureFlyctlAvailable` and `flyctlBuildAndPush` — with no corresponding test file.

The project's established pattern is `src/lib/docker.test.ts`, which covers **14 distinct cases** across `ensureDockerAvailable`, `dockerBuild`, `dockerLogin`, and `dockerPush`:
- binary available → no throw
- binary missing (ENOENT) → `CLIError` with install instructions
- daemon down / non-zero exit → `CLIError` with stderr detail
- token/secret never appears in argv
- etc.

The new `flyctl.ts` module has analogous code paths — `r.error`, `r.status !== 0`, stderr slice — that need the same treatment. A `flyctl.test.ts` modelled on `docker.test.ts` is required before merge. At minimum:

| Case | Function |
|---|---|
| `flyctl version` exits 0 → no throw | `ensureFlyctlAvailable` |
| `flyctl version` returns ENOENT → `CLIError` mentioning install URL and `--image` | `ensureFlyctlAvailable` |
| `flyctl version` exits non-zero → `CLIError` with stderr detail | `ensureFlyctlAvailable` |
| successful build → returns `{ imageRef }` with correct `registry.fly.io/<appId>:<imageLabel>` | `flyctlBuildAndPush` |
| `flyctl deploy` spawn fails (error) → `CLIError` with message | `flyctlBuildAndPush` |
| `flyctl deploy` exits non-zero → `CLIError` mentioning exit code | `flyctlBuildAndPush` |
| `token` is passed via `FLY_API_TOKEN` env var, NOT in argv | `flyctlBuildAndPush` (security invariant) |

The insforge-development skill states: *"Every new behavior has at least one test"* and *"Every bug fix has a regression test."* The `docker.test.ts` proves this is taken seriously here.

---

### Suggestion

#### [Functionality] Misleading JSDoc comment about fly.toml handling — `src/lib/flyctl.ts:43-45`

The docstring says:

> *"If a fly.toml exists we append `app = "<appId>"` via env var so it doesn't conflict with whatever the user wrote there."*

The actual code passes `--app opts.appId` as a CLI flag, not via any env var. The `--app` flag does correctly override fly.toml app name, so the behavior is right — but the comment describes a mechanism that does not exist and would mislead anyone debugging a conflict. Update the comment to reflect what the code actually does.

#### [Software Engineering] No source-mode integration test stub

`src/integration/compute.test.ts` tests image-mode deploy (line 32) but has no test — even a skipped/stub one — for source-mode deploy. The PR test plan calls out a manual smoke-test against staging, but given that image-mode already has an integration test under `INTEGRATION_TEST_ENABLED`, it would be consistent to add a source-mode counterpart (even if it's `it.skip`ed until the companion cloud PR merges). Not a blocker, but worth addressing before this feature is considered fully covered.

#### [Functionality] `imageRef` returned without verification against flyctl output — `src/lib/flyctl.ts:50, 77`

`flyctlBuildAndPush` constructs `imageRef` from inputs (`registry.fly.io/${opts.appId}:${opts.imageLabel}`) and returns it as ground truth. Since `stdio: 'inherit'` suppresses captured output, there is no parse-and-validate step confirming that flyctl actually pushed to that exact ref. If an older flyctl version silently normalises the image label (e.g., strips characters, truncates), step 4 of the deploy (`PATCH imageUrl`) would reference a non-existent image. Either document the minimum required flyctl version that guarantees `--image-label` is honoured verbatim, or add a version-check / assertion.

---

### Information

#### `imageLabel = \`cli-${Date.now()}\`` — `src/commands/compute/deploy.ts:199`

Two rapid concurrent deploys of the same service from the same machine could generate the same millisecond timestamp and collide in the registry tag. A short random suffix (e.g., `cli-${Date.now()}-${Math.random().toString(36).slice(2,7)}`) would eliminate the race for free.

#### No security-relevant changes beyond token-env handling

The token is correctly placed in `FLY_API_TOKEN` (env), not in argv, so it won't appear in `ps` output. The existing API token for the InsForge backend continues to flow via credentials as before. No new secrets-logging risk identified.

---

## Verdict

`request_changes` — one Critical finding (missing unit tests for `src/lib/flyctl.ts`). Everything else is non-blocking.

jwfing (approved):
LGTM, approved.

PR #91, opened 2026-04-29
fix(cli/compute): write fly.toml stub on fresh apps + roll back on build failure

tonychang04:
## Summary

Fixes two source-mode deploy bugs in 0.1.60 that block first-time users (closes the issue caught while testing 0.1.60 today). Bumps to **0.1.61**.

### Bug 1: Missing fly.toml stub → 100% failure on first deploy

```
$ npx @insforge/cli compute deploy . --name my-api --port 8080
Detected Dockerfile at .../Dockerfile
Creating service "my-api"...
Created Fly app my-api-<projectId>
Requesting deploy token...
Building & pushing on Fly remote builder...
==> Verifying app config
Error: failed to grab app config from existing machines, error: could not create a fly.toml from any machines :-(
No machines configured for this app
Error: flyctl deploy --build-only failed (exit 1).
```

`flyctl deploy --remote-only` derives app config from existing machines; on a brand-new Fly app there are zero machines to derive from, so flyctl errors out. The old code comment claimed "flyctl will invent a fly.toml if none exists" — that was wrong.

**Fix:** `ensureFlyTomlStub()` writes a minimal `fly.toml` (`app`, `primary_region`, `internal_port` + `http_service` block) before invoking flyctl, and removes it on exit. If the user already has a `fly.toml`, we leave it alone — power users keep full override.

### Bug 2: Zombie service on build failure

Before the fix, any build failure (which was 100% on first deploy because of Bug 1) left a half-created service: DB row + empty Fly app, `flyMachineId: null`, status `deploying`. The user had to manually `compute delete` it before retrying.

**Fix:** `deploy.ts` catches build failure on a freshly-created service and DELETEs the row through the OSS proxy (which also destroys the Fly app). For redeploys of an existing service we **don't** roll back — the running machine should survive a transient build error.

## Design choice: auto-stub vs. user-provided fly.toml

The fix gives **both**, in priority order:

1. **User-provided `fly.toml`** (advanced) — fully respected; CLI never touches it
2. **Auto-stub** (default) — written transparently for the deploy duration, deleted after; user never sees it

Telling users to write their own `fly.toml` would push a flyctl-specific quirk into the user's workflow when they're just trying to ship a Dockerfile. The stub is a 30-line workaround in the CLI — invisible to 95% of users, while leaving the door open for the 5% who need custom fly.toml config.

## Test plan

**Unit tests** (`src/lib/flyctl.test.ts`, +5 new tests, 16/16 pass):
- Stub written when no fly.toml exists, with correct `app` / `primary_region` / `internal_port`
- Stub left alone when user-provided fly.toml exists
- Stub cleaned up on success, non-zero exit, AND spawn error

**Live e2e on staging today** (`api-beta.insforge.dev` cloud `v1.2.11-compute2` + OSS `v2.1.3-compute`):

| Scenario | Result |
|---|---|
| Fresh dir, only Dockerfile + index.html, no fly.toml | ✓ deployed `[running]`, live HTTP 200, no leftover `fly.toml` after deploy |
| Redeploy to existing service (in-place update) | ✓ updated `[running]`, no rollback |
| Broken Dockerfile (`RUN false`) | ✓ flyctl exits 1 → `Rolled back service "broken-..." after build failure` → 0 zombies in `compute list` |

## Diff stats

```
src/commands/compute/deploy.ts  | +20  (rollback path on freshly-created service)
src/lib/flyctl.ts               | +50  (ensureFlyTomlStub + plumb region/port)
src/lib/flyctl.test.ts          | +75 (new file: 16 tests, +5 vs the WIP)
package.json                    | 0.1.60 → 0.1.61
package-lock.json               | sync
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved deployment resilience: failed remote build/push now triggers rollback for freshly created services and emits success/failure messages; existing services skip rollback.
  * Build/push failures continue to surface so deployments still fail when appropriate.

* **New Features**
  * Deploy now respects explicit region and port settings and creates a minimal config stub when missing (auto-cleaned up).

* **Tests**
  * Added thorough tests covering CLI availability, build/push behavior, and stub lifecycle.


jwfing (approved):
LGTM, approved.

PR #98, opened 2026-05-01
feat(cli): --auth flag for overlaying auth-provider scaffolds

tonychang04:
## Summary

Adds a `--auth <provider>` flag to `create` and `link` that overlays a third-party auth integration onto whatever the user has — fresh project, blank dir, or existing app. Today: `--auth better-auth`. Same shape extends to clerk/auth0/workos/kinde/stytch.

```bash
# Bolt BA onto your existing Next.js app:
npx @insforge/cli link --auth better-auth

# Or start fresh with nextjs + BA wired:
npx @insforge/cli create --auth better-auth
npx @insforge/cli link --template nextjs --auth better-auth
```

## Architecture: scaffolds live in the templates repo, fetched at runtime

The earlier draft of this PR bundled scaffolds into the CLI and shipped them via a `tsup` copy hook. After review (and a second look at how this scales to clerk/auth0/etc.), the scaffolds were moved into the [InsForge/insforge-templates](https://github.com/InsForge/insforge-templates) repo under `auth-providers/<name>/`, alongside a declarative `manifest.json`. The CLI no longer ships scaffold files at all.

At runtime, `applyAuthProvider()`:

1. Shallow-clones the templates repo to `tmpdir()` (via `execFile` with argv form, not `exec` — env vars `INSFORGE_TEMPLATES_REPO` / `INSFORGE_TEMPLATES_BRANCH` are validated against a safe-character pattern before they ever reach git's argv)
2. Reads `auth-providers/<provider>/manifest.json` to get the file list, package.json patch, .env block, and next-steps
3. Runs the overlay engine (file copy, package.json deep-merge, env-collision overlay)
4. Cleans up the tempdir

Why this shape:
- One repo per provider on the templates side — no need to release a CLI version every time we add or tweak a provider
- `auth-providers/` lives in the templates repo but is *invisible to the `npx @insforge/cli create` template picker* — it's only reachable via `--auth`, so it doesn't pollute the picker as we add more providers
- Each new auth provider = one directory + manifest in the templates repo, zero CLI changes
- The CLI bundle is leaner — `dist/` no longer carries 13 raw scaffold assets

The corresponding templates-repo PR ([insforge-templates#25](https://github.com/InsForge/insforge-templates/pull/25)) is merged to `main`.

## Overlay rules (existing-project-safe)

Every operation is additive — `--auth` never clobbers user code:

| File | Behavior |
|---|---|
| `lib/auth.ts`, `lib/insforge*.ts`, route handlers, sql/, scripts/ | Copy if not exists, skip + warn if exists |
| `package.json` | Deep-merge dependencies/devDependencies/scripts. **User's existing values win** — `@insforge/sdk: ^1.5.0` is NOT downgraded to ^1.2.6, `scripts.dev: next dev --port 4000` is NOT clobbered with `next dev` |
| `.env.example` | Appended (BA section after user's vars). Per-key collision filter: any KEY already present is silently dropped from the appended block (and reported via `envKeysSkipped`) |
| `.env.local` | Created with auto-filled `INSFORGE_*_BASE_URL` (project oss_host), `INSFORGE_*_ANON_KEY` (fetched), `INSFORGE_JWT_SECRET` (fetched from `/api/secrets/JWT_SECRET`), `BETTER_AUTH_SECRET` (32 random hex). Same per-key collision filter when the file already exists |

User-supplied: `DATABASE_URL` (BA needs Postgres for its own tables; self-hosted has different defaults than cloud — left explicit).

User-runs: `npm install && npm run setup && npm run dev`. `npm run setup` chains `auth:migrate` (BA CLI creates user/session/account/verification tables) + `db:setup` (Node+pg runner applies sql/01-init.sql + sql/02-revoke.sql for RLS helper, notes table, and REVOKE block).

## Tested

End-to-end against the local `ba-sdk-test` stack (Postgres :5433, OSS :7230) on the merged templates `main`:
- `git clone --depth 1` from `InsForge/insforge-templates.git` → 13 scaffold files written, 0 skipped
- package.json deep-merge — user's `@insforge/sdk: ^1.5.0` preserved over manifest's `^1.2.6`
- `.env.example` collision filter — user's `MY_OWN_VAR` and any pre-existing BA keys preserved
- `.env.local` synthesized with real anon key + JWT secret fetched from the local stack
- `npm install && npm run setup && next dev` → signup → bridge JWT → RLS-scoped INSERT → cross-user isolation verified → REVOKE block returns 42501 for direct PostgREST access to `user`/`session`/`account`/`verification`

Plus, after this push: `INSFORGE_TEMPLATES_BRANCH='main; rm -rf …'` is rejected by the safe-character guard before git runs.

**CLI build:** clean. `dist/index.js` 277.08 KB. `npm run lint` and `vitest run` (156 tests) pass.

## Test plan

- [ ] `node dist/index.js link --help` shows the `--auth` flag
- [ ] `link --auth bogus ...` → "Invalid --auth bogus"
- [ ] `link --template nextjs --auth better-auth` against a real linked project → fresh subdir with full nextjs+BA wiring
- [ ] `link --auth better-auth` (no --template) against an existing Next.js project → BA overlay, no clobber
- [ ] After link: `npm install && npm run setup && npm run dev` works end-to-end
- [ ] `npx @insforge/cli create` interactive picker still shows the same 6 frontend options (no auth surfacing)
- [ ] `--auth` failure under `--json` → primary success JSON on stdout stays parseable; `{"warning": "..."}` line lands on stderr


jwfing (changes requested):
## Summary

The overlay engine and env-collision logic are solid, but there is a critical architecture mismatch between what the PR description claims was built and what was actually shipped — combined with a command-injection risk introduced in both the new and existing template-clone paths.

---

## Requirements context

No `/docs/superpowers/` directory exists in this repo. Review is assessed against the PR description and code alone.

---

## Findings

### Critical

#### Functionality — Architecture mismatch / feature likely broken for all users

**`tsup.config.ts`, `src/auth-providers/apply.ts:120–158`, `src/auth-providers/` directory**

The PR description states:

> _"Auth providers live inside the CLI, not in the templates repo … shipped to `dist/auth-providers/<name>/files/` via a `tsup` `onSuccess` copy hook … CLI build: clean. `dist/auth-providers/better-auth/files/` has all 10 raw assets after `npm run build`."_

None of this is true in the submitted code:

1. **`tsup.config.ts` has no `onSuccess` hook** — and `clean: true` means every build wipes `dist/`. The claimed `dist/auth-providers/better-auth/files/` cannot exist after a clean build.
2. **`src/auth-providers/` contains only `apply.ts` and `apply.test.ts`** — there is no `better-auth/` subdirectory, no `manifest.ts`, and no `files/` assets anywhere in the CLI repo.
3. **The actual implementation still clones from `insforge-templates.git` at runtime** (`apply.ts:138–143`). The code comment at the very top of `apply.ts` even says: _"Auth-provider scaffolds live in the InsForge templates repo under `auth-providers/<provider>/`"_ — directly contradicting the PR description.

The PR description also says: _"The original insforge-templates#24 is closed; everything lives here."_ But "here" (CLI) doesn't have the scaffold files. If `auth-providers/better-auth/` doesn't exist in the templates repo (because that PR was cancelled), **every user running `--auth better-auth` will immediately hit the error: `Auth provider "better-auth" not found in templates repo`**.

This needs to be resolved before merge: either (a) complete the bundled approach (add asset files + tsup copy hook), or (b) confirm there is a companion templates-repo PR with the scaffold and update the PR description to reflect the real "clones from templates repo" architecture.

---

### Suggestion

#### Security — Shell injection via `INSFORGE_TEMPLATES_BRANCH` / `INSFORGE_TEMPLATES_REPO`

**`src/auth-providers/apply.ts:138–143`, `src/commands/create.ts:590–596`**

Both `fetchProviderTree` (new) and `downloadGitHubTemplate` (modified) interpolate these env vars directly into a shell command string:

```ts
const branchFlag = branch ? ` -b ${branch}` : '';
await execAsync(`git clone --depth 1${branchFlag} ${repo} .`, ...);
```

A value like `main; rm -rf /` or a repo URL containing shell metacharacters triggers command injection. These are developer-facing escape hatches, but they could be set in shared CI environments or injected via `.env` files. Use `execFile` (or `spawn`) with an explicit args array instead:

```ts
const args = ['clone', '--depth', '1', ...(branch ? ['-b', branch] : []), repo, '.'];
await execFileAsync('git', args, { cwd: tempDir, maxBuffer: 10 * 1024 * 1024, timeout: 60_000 });
```

Note: this pattern was introduced in both the new `apply.ts` and in the modified `create.ts` `downloadGitHubTemplate` path.

---

#### Functionality — `--auth` errors silently swallowed in `--json` mode

**`src/commands/create.ts:385–394`, `src/commands/projects/link.ts:153–159`, `link.ts:200–209`, `link.ts:376–385`, `link.ts:420–430`**

Every `applyAuthProvider` call site wraps the call in `try/catch` and only logs a warning when `!json`. When `json: true` (machine-readable / CI mode), auth failures produce **no output and no non-zero exit code**. A consumer piping JSON output has no way to detect that auth wiring failed. At minimum, include an `auth_error` field in the JSON output on failure.

---

#### Functionality — `result.nextSteps` not displayed in `create` or `link --template --auth`

**`src/commands/create.ts:386–390`, `src/commands/projects/link.ts:153–159`, `link.ts:376–383`**

The `ApplyResult.nextSteps` string (e.g. "Run npm install && npm run setup && npm run dev") is only shown in the two `link --auth` (no-template) paths. It is silently dropped in `create --auth` and `link --template --auth`. Users who use the template+auth combo won't see the post-setup instructions.

---

#### Software engineering — Test coverage gaps

**`src/auth-providers/apply.test.ts`**

The tests cover only the two exported utility functions (`extractEnvKeys`, `filterCollidingEnvLines`). Missing:

- A test for the `--auth bogus` validation path in `applyAuthProvider` (the guard at `apply.ts:175`)
- Any test exercising the overlay logic (file-copy skip-if-exists, `deepMergeKeepBase`, `.env.example` collision resolution)
- `deepMergeKeepBase` has no unit tests despite being a non-trivial recursive merge with specific base-wins semantics

The PR description claims extensive manual overlay testing, but these scenarios are unencoded as regression tests.

---

### Information

#### Code comment at top of `apply.ts` contradicts PR description

**`src/auth-providers/apply.ts:1–12`**

The module-level comment correctly describes the implementation: _"Auth-provider scaffolds live in the InsForge templates repo"_. The PR description claims the opposite. Once the Critical finding above is resolved, this comment should either stay as-is (if the templates-repo architecture is kept) or be updated (if the bundled approach is implemented).

---

## Verdict

`request_changes` — one Critical finding (architecture mismatch + likely runtime failure for all users). The security and silent-failure items above are non-blocking but should be addressed before this ships widely.

jwfing (changes requested):
## Summary

Solid architecture overall — the manifest-driven, runtime-fetch approach scales cleanly to new providers and the overlay-safe file/env logic is correct and well-described. One security hole must be closed before merge.

## Requirements context

No `/docs/superpowers/` directory exists in this repo. Assessment is based on the PR title, body, and the InsForge development conventions documented in `.claude/skills/insforge-development/SKILL.md`.

---

## Findings

### Critical

#### Security — Shell injection in `downloadGitHubTemplate` (`src/commands/create.ts:595-600`)

The PR introduces `INSFORGE_TEMPLATES_REPO` and `INSFORGE_TEMPLATES_BRANCH` env-var support into `downloadGitHubTemplate`, but **does not apply the same guards** that `fetchProviderTree` in `apply.ts` correctly uses:

```typescript
// apply.ts — CORRECT: argv form, validated first
const SAFE_REPO_PATTERN = /^(https?:\/\/|git@)[A-Za-z0-9._:/@~+-]+(\.git)?$/;
const SAFE_BRANCH_PATTERN = /^[A-Za-z0-9._/-]+$/;
// … validation throws if pattern fails …
await execFileAsync('git', args, { cwd: tempDir, … });   // no shell

// create.ts — VULNERABLE: shell string, no validation
const branchFlag = templatesBranch ? ` -b ${templatesBranch}` : '';
await execAsync(
  `git clone --depth 1${branchFlag} ${templatesRepo} .`,  // shell-interpolated
  { cwd: tempDir, … },
);
```

`execAsync` wraps `child_process.exec`, which passes the command through `/bin/sh`. An attacker who can influence `INSFORGE_TEMPLATES_BRANCH` (e.g., via a `.env` file in a CI pipeline, a compromised dev machine, or a supply-chain attack on an env loader) gets arbitrary command execution. The PR description explicitly states "INSFORGE_TEMPLATES_BRANCH='main; rm -rf …' is rejected by the safe-character guard" — but that is only true for `applyAuthProvider`. The `create --template` and `link --template` flows (which both call `downloadGitHubTemplate`) are not guarded.

**Fix:** Apply the same patterns from `apply.ts` and switch to `execFileAsync` with an argv array — identical to what `fetchProviderTree` already does.

---

### Suggestion

#### Software Engineering — Thin test coverage for `applyAuthProvider` (`src/auth-providers/apply.test.ts`)

Only the two pure helper functions (`extractEnvKeys`, `filterCollidingEnvLines`) are unit-tested. The main orchestration function `applyAuthProvider`, and the helpers `deepMergeKeepBase`, `walkFiles`, and `loadManifest`, have no automated coverage. These are the most complex and failure-prone parts: the package.json merge, the `.env.local` synthesise-and-extend path, and the "skip-if-exists" file copy. A simple fs-mock or temp-directory fixture would cover the merge/overlay logic without requiring a network call or a real git clone (the `fetchProviderTree` call can be injected or the env var overridden in tests).

The PR says testing was done manually end-to-end; that's valuable, but it means regressions in the overlay logic won't be caught automatically.

#### Functionality — No manifest shape validation in `loadManifest` (`src/auth-providers/apply.ts:145-150`)

```typescript
return JSON.parse(await fs.readFile(manifestPath, 'utf-8')) as AuthProviderManifest;
```

This is a blind cast. If the templates repo ships a malformed manifest (missing `files`, `nextSteps`, or `envExampleAppend`), the failure will surface as a cryptic `TypeError: Cannot read properties of undefined` buried somewhere inside `applyAuthProvider`, not as a helpful "invalid manifest" error. A minimal check (`if (!Array.isArray(manifest.files) || …) throw new Error(…)`) would make debugging far easier when the templates and CLI repos diverge.

#### Functionality — Inconsistent JSON error handling for auth failures in `link.ts` (`src/commands/projects/link.ts:155`)

One of the four auth-failure catch blocks uses:
```typescript
if (json) outputJson({ success: false, warning: msg });   // stdout, implies command failure
```
while the other three use:
```typescript
if (json) console.error(JSON.stringify({ warning: msg }));  // stderr, non-fatal
```
Since `--auth` failures are explicitly designed to be non-fatal warnings (the link succeeded; only the scaffold overlay failed), the `outputJson({ success: false, … })` call will confuse automation tooling that checks stdout for `success: false`. It should use `console.error(JSON.stringify({ warning: msg }))` for consistency with the other three sites.

---

### Information

*(none)*

---

## Verdict

`request_changes` — one Critical finding (shell injection in `downloadGitHubTemplate`) must be fixed. The Suggestion items are worth addressing before the next auth provider ships, but are not blocking on their own.

tonychang04:
Thanks for the careful read — pushed `bafe26a` addressing all four findings.

## Critical: shell injection in `downloadGitHubTemplate` ✅

Mirrored the exact pattern from `fetchProviderTree`:
- `exec` → `execFile` (argv form, no shell)
- `INSFORGE_TEMPLATES_REPO` validated against `SAFE_REPO_PATTERN`
- `INSFORGE_TEMPLATES_BRANCH` validated against `SAFE_BRANCH_PATTERN`
- `--` separator before the repo URL so even an unblocked value can't be reinterpreted as an option

Confirmed end-to-end against the rebuilt CLI:

```bash
$ INSFORGE_TEMPLATES_BRANCH='main; rm -rf /tmp/template-injection-test' \
    node dist/index.js link --template nextjs --api-base-url http://localhost:7230 \
    --api-key fake-key --json

# stdout — primary success JSON, parseable
{ "success": true, "project": {...}, "directory": "insforge-nextjs", "template": "nextjs" }

# stderr — structured warning
{"warning":"Failed to download nextjs template: INSFORGE_TEMPLATES_BRANCH has unsupported characters: main; rm -rf /tmp/template-injection-test"}

# target dir intact, no rm -rf executed
```

Bonus: while in there, made the JSON-mode failure surface on stderr instead of being silently swallowed (parallel to the `--auth` path you already flagged).

## Suggestion: inconsistent JSON error handling in link.ts:159 ✅

Caught — that one site survived my earlier `replace_all` because it had a slightly different "before" string. Fixed; all four catch blocks now use `console.error(JSON.stringify({ warning: msg }))`.

## Suggestion: manifest shape validation in `loadManifest` ✅

Replaced the blind cast with a field-by-field check:

```typescript
const missing: string[] = [];
if (typeof parsed.name !== 'string') missing.push('name');
if (!Array.isArray(parsed.files)) missing.push('files (array)');
if (!parsed.packageJsonPatch || typeof parsed.packageJsonPatch !== 'object') missing.push('packageJsonPatch (object)');
if (typeof parsed.envExampleAppend !== 'string') missing.push('envExampleAppend (string)');
if (typeof parsed.nextSteps !== 'string') missing.push('nextSteps (string)');
if (missing.length > 0) {
  throw new Error(`Malformed manifest.json in ${providerDir} — missing or wrong-typed fields: ${missing.join(', ')}`);
}
```

So a templates/CLI version skew now fails with `Malformed manifest.json … missing or wrong-typed fields: nextSteps (string)` instead of `Cannot read properties of undefined`.

## Suggestion: thin test coverage for `applyAuthProvider`

Acknowledged but deferring to a follow-up. The change here is already merging weight, and the integration test you're describing (fs-mock or temp-dir fixture for `applyAuthProvider` + `deepMergeKeepBase` + `walkFiles` + `loadManifest`) is a self-contained unit of work — better as its own PR than tacked onto a security fix. I'll open it as a follow-up before the next provider lands.

## Verification after this push

- `npm run build` clean (278.27 KB)
- `npx eslint src/` no warnings
- `npx vitest run` 156 passed / 13 skipped
- E2E against live local stack still green: signup → bridge → RLS-scoped INSERT → cross-user isolation → REVOKE 42501 on user/session/account/verification

PTAL.

jwfing (commented):
## Code Review — `feat(cli): --auth flag for overlaying auth-provider scaffolds`

**Summary:** Solid feature addition. The runtime-fetch architecture (scaffolds in the templates repo, validated argv-form git clone, per-key collision filter) is well-designed. No blocking issues found.

---

### Requirements context

No `/docs/superpowers/` directory exists in this repo — assessment based on the PR description and linked `insforge-templates#25` (already merged to `main`).

---

### Findings

#### Critical
_(none)_

---

#### Suggestion

**[S1 — Software Engineering] Overlay semantics contradict the PR description, and `skipped` is dead code**
`apply.ts:74` (`skipped: []`) and `apply.ts:189–201`

The PR body's overlay table says:
> `lib/auth.ts`, route handlers, … → _Copy if not exists, skip + warn if exists_

But the code comment at `apply.ts:189` says the opposite:
> _Auth-provider overlays REPLACE on collision — the user opted into `--auth <provider>` expressing intent that the provider's auth primitives win._

…and the code unconditionally calls `fs.copyFile` for every file. The `skipped: string[]` field in `ApplyResult` is initialized to `[]` and never populated — it is dead code. The replace-on-collision rationale in the code comment is sound (avoids silently broken exports from base templates), but the PR description table should be updated to match, and the `skipped` field should be removed or used.

---

**[S2 — Functionality] `create --auth` never shows `result.nextSteps`**
`src/commands/create.ts:394–405` vs `src/commands/projects/link.ts:226`

`link.ts` shows `clack.note(result.nextSteps, "What's next")` after the overlay; `create.ts` only logs the file-count success message and discards `result.nextSteps`. Users who run `create --auth better-auth` won't see the setup instructions (`npm install && npm run setup && npm run dev`).

---

**[S3 — Software Engineering] Duplicated regex constants**
`src/auth-providers/apply.ts:116–117` and `src/commands/create.ts:29–30`

`SAFE_REPO_PATTERN` and `SAFE_BRANCH_PATTERN` are copy-pasted verbatim into two files. If one changes (e.g., to add `ssh://` support), the other silently drifts. Extract both to a shared `src/lib/git-safety.ts` (or similar).

---

**[S4 — Software Engineering] Brittle idempotency guard in `.env.example` append**
`src/auth-providers/apply.ts:230–231`

```ts
if (!existing.includes('# ─── Better Auth + InsForge bridge')) {
```

This hardcodes a comment string from the templates repo's `manifest.json`. If that comment is ever reworded in the templates repo, the CLI's idempotency check silently breaks and a second `--auth` run would double-append the block. Consider a more stable sentinel (e.g., `# @insforge-auth better-auth`) or a dedicated marker key (e.g., `_INSFORGE_AUTH_APPLIED=better-auth`) that the CLI itself writes.

---

**[S5 — Software Engineering] Thin test coverage for the overlay engine**
`src/auth-providers/apply.test.ts`

47 lines test only the two pure utility functions (`extractEnvKeys`, `filterCollidingEnvLines`). The core overlay logic — `deepMergeKeepBase`, package.json creation for fresh projects (the `else` branch at `apply.ts:253–261`), and file-copy overwrites — has no automated tests. A temp-dir integration test for `applyAuthProvider` would catch regressions in the main path (the PR description's e2e story is compelling, but that's manual). Even a focused unit test for `deepMergeKeepBase` would be valuable given the "base wins" invariant is a key correctness guarantee.

---

#### Information

**[I1 — Software Engineering] Stale comment in `link.ts`**
`src/commands/projects/link.ts:47–50`

```ts
// overlaying an auth-provider scaffold from CLI's bundled assets —
// no templates-repo directory required.
```

The implementation fetches from the templates repo at runtime, not from bundled assets. This comment is a leftover from the earlier draft mentioned in the PR body.

---

**[I2 — Performance] Two sequential git shallow-clones when `--template` and `--auth` are both passed**
`src/commands/projects/link.ts:135` + `src/auth-providers/apply.ts:166`

Both `downloadGitHubTemplate` and `fetchProviderTree` shallow-clone `InsForge/insforge-templates.git` independently. With `link --template nextjs --auth better-auth` users pay two round-trips to GitHub (~2–5 s each on a typical connection). Not blocking now (single-provider MVP), but worth a one-clone refactor before adding clerk/auth0/etc.

---

**[I3 — Software Engineering] `npm install` uses shell (`execAsync`) while git uses `execFileAsync`**
`src/commands/projects/link.ts:218`

No injection risk — the command is a hardcoded string with no user data — but it is inconsistent with the PR's own hardening stance on argv-form execution. Low priority.

---

**[I4 — Functionality] `npm install` silently skipped in `--json` mode after patch**
`src/commands/projects/link.ts:214`

When `result.packageJsonPatched && !json` is false because `json === true`, the deps are added to `package.json` but never installed. This is intentional for scripting contexts, but the emitted JSON output gives no hint that `npm install` is needed. Consider including `"requiresInstall": true` in the JSON output.

---

### Verdict

**approved** (informational — no Critical findings). The security hardening (`execFile`, argv validation, env-key collision filter, `randomBytes` for secrets) is thorough and the architecture is cleanly extensible. Address S1–S5 at your discretion; I1–I4 are take-or-leave.

jwfing (approved):
LGTM, approved.

jwfing (approved):
LGTM, approved.

PR #105, opened 2026-05-07
feat(link): auto-fill DATABASE_URL from cloud project (+ bump 0.1.67)

tonychang04:
Cloud projects expose their Postgres URL via
`/api/metadata/database-connection-string`, but the BA scaffold's `.env.local` was leaving DATABASE_URL at the localhost default — cloud users had to fetch and paste the URL manually before `npm run setup` would work.

Add a `getDatabaseConnectionString()` helper alongside the existing `getJwtSecret()` / `getAnonKey()`, and use it in `applyAuthProvider` so DATABASE_URL gets the same pre-fill treatment as the other env vars. Self-hosted projects keep the localhost default (the helper returns null when the cloud endpoint 500s on missing PROJECT_ID).

Pairs with insforge-templates#31 which fixes the SSL cert error that the cloud connection string would otherwise trigger.

Bump 0.1.66 → 0.1.67.

## Summary by CodeRabbit

* **New Features**
  * Database connection string is now automatically fetched and configured during authentication provider setup.

* **Chores**
  * Version updated to v0.1.67.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@insforge/cli",
-  "version": "0.1.66",
+  "version": "0.1.67",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@insforge/cli",
-      "version": "0.1.66",
+      "version": "0.1.67",
       "license": "Apache-2.0",
       "dependencies": {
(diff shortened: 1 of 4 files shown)


jwfing (approved):
LGTM, approved.

PR #106, opened 2026-05-07
fix(link): refresh stale platform defaults in .env.local on re-link

tonychang04:
## Summary

When \`link --auth <provider>\` ran against a directory that already had \`.env.local\`, the auto-fill skipped every key the user already defined — even if the user's value was just the **manifest's literal default from a prior run**. Cloud users who re-linked after the DATABASE_URL auto-fill landed (#105) kept their old localhost URL, hit \"🚀 No migrations needed\" against the wrong DB, and got silently broken setups.

## Repro

\`\`\`bash
mkdir test && cd test && npm init -y
# scaffold once with an old CLI (or while only LOCALHOST DATABASE_URL exists in manifest)
npx @insforge/cli@0.1.66 link --project-id <cloud> --auth better-auth
# .env.local now has DATABASE_URL=postgresql://postgres:postgres@127.0.0.1:5432/insforge

# upgrade + re-link
npx @insforge/cli@latest link --project-id <cloud> --auth better-auth
# .env.local still has the localhost URL — the auto-fill skipped it because
# the key already existed.
\`\`\`

## Fix

Compare the user's existing value against the literal default in \`manifest.envExampleAppend\`. If they match AND the platform now has a real value, overwrite the line in-place. User-customized values stay untouched.

\`\`\`
fresh .env.local                                                  → unchanged (writes filled blob, like today)
user customized X                                                 → unchanged (stays user's, like today)
user has manifest default for X, platform has real value          → REFRESHED ✨
user has manifest default for X, platform has no value (self-host)→ unchanged (no-op)
key not in user's file                                            → unchanged (appended via existing collision flow)
\`\`\`

Adds:
- \`extractEnvPairs(content)\` — KEY → value Map (counterpart to existing \`extractEnvKeys\`)
- \`refreshStaleEnvDefaults(existing, manifestDefaults, platformValues)\` — line-level in-place rewrite
- \`ApplyResult.envKeysRefreshed\` — surfaces a \"Refreshed stale platform defaults: DATABASE_URL\" info line so the user can tell what changed

## Bump

0.1.67 → 0.1.68

## Test plan

- [x] build passes (\`npm run build\`)
- [x] lint 0 errors (\`npm run lint\`)
- [ ] (post-merge) e2e: re-link an existing dir against a cloud project, verify DATABASE_URL gets refreshed to the cloud URL while user-customized vars stay

## New Features
- Environment variable management improvements: intelligently refreshes stale defaults from the platform while preserving user customizations in `.env.local`
- Added tracking and logging of refreshed environment configuration keys during setup

## Chores
- Version bumped to 0.1.68

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@insforge/cli",
-  "version": "0.1.67",
+  "version": "0.1.68",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@insforge/cli",
-      "version": "0.1.67",
+      "version": "0.1.68",
       "license": "Apache-2.0",
       "dependencies": {
         "@clack/prompts": "^0.9.1",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@insforge/cli",
-  "version": "0.1.67",
+  "version": "0.1.68",
   "description": "InsForge CLI - Command line tool for InsForge platform",
   "type": "module",
   "bin": {
--- a/src/auth-providers/apply.ts
+++ b/src/auth-providers/apply.ts
@@ -74,6 +74,7 @@ interface ApplyResult {
   envExampleAppended: boolean;
   envLocalWritten: boolean;
   envKeysSkipped: string[];
+  envKeysRefreshed: string[];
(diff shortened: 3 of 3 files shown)


jwfing (approved):
LGTM, approved.

PR #107, opened 2026-05-07
test(apply): cover refreshStaleEnvDefaults regression suite

tonychang04:
## Summary

Adds 13 unit tests for the env-merge logic that 0.1.68 introduced. Acts as regression guard against the fix — if anyone changes the matching/replace logic, tests will catch behavioral drift.

## Coverage

```
extractEnvPairs:
  ✓ returns KEY → value pairs
  ✓ preserves URL-style values verbatim
refreshStaleEnvDefaults:
  ✓ replaces user value when it matches manifest default and platform has real value
  ✓ preserves user value when it differs from the manifest default
  ✓ skips refresh when platform has no real value (self-hosted, helper returned null)
  ✓ handles multiple keys, refreshing only the stale ones
```

All 13 pass locally (`npx vitest run src/auth-providers/apply.test.ts` → `Tests 13 passed (13)`).

---
## Summary by cubic
Add 13 unit tests for the env-merge path introduced in 0.1.68, covering `extractEnvPairs` and `refreshStaleEnvDefaults`. Tests ensure we only replace values that still equal the manifest default when the platform has a real value, keep user customizations, and handle multiple keys.

<sup>Written for commit fd0f80a70cfda8703d09807f70a65f43eccd849f. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Tests**
  * Expanded test coverage for environment variable handling and platform value synchronization functionality.

### DIFF
--- a/src/auth-providers/apply.test.ts
+++ b/src/auth-providers/apply.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest';
-import { extractEnvKeys, filterCollidingEnvLines } from './apply.js';
+import { extractEnvKeys, extractEnvPairs, filterCollidingEnvLines, refreshStaleEnvDefaults } from './apply.js';
 
 describe('extractEnvKeys', () => {
   it('finds plain KEY=value lines', () => {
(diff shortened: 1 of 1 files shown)


tonychang04:
Content already present in main (verified via diff: PR's 13 added tests are all in main as of latest rebase; only delta is one extra test main has on top). Closing as superseded.

PR #112, opened 2026-05-08
fix switch bug(oss host), and optimise branch create flow.

jwfing:
---
## Summary by cubic
Fixes branch switch writing a bare `oss_host` and auto-heals existing configs. Improves branch create with a single `@clack/prompts` spinner covering create, provisioning, and the auto-switch, with clear switch‑failure messaging; JSON mode unchanged.

- **Bug Fixes**
  - Switch saves `oss_host` via `buildOssHost()` (https); `getProjectConfig()` normalizes legacy bare-host values on read.
  - After a successful create, a switch error is reported as “switching context failed” with a retry hint, not “creation failed”.

- **New Features**
  - One spinner spans create → provisioning → switch, stays active through switch, and failures render in the same frame; `runBranchSwitch` now runs with `silent: true`.
  - JSON mode still outputs only the final `{ branch }` payload.

<sup>Written for commit 34f5383bcc5d3645aa9211258372a8ea2bec29d2. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Spinner feedback now spans branch creation, provisioning, and optional auto-switch for clearer CLI progress.

* **Bug Fixes**
  * Legacy OSS host values are normalized to a valid HTTPS URL; OSS host construction centralized to ensure consistent URLs.

* **Chores**
  * Bumped package version to 0.1.70.

* **Tests**
  * Added tests for spinner lifecycle and OSS host construction; updated branch-switch tests to reflect URL behavior.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@insforge/cli",
-  "version": "0.1.69",
+  "version": "0.1.70",
   "description": "InsForge CLI - Command line tool for InsForge platform",
   "type": "module",
   "bin": {
--- a/src/commands/branch/create.test.ts
+++ b/src/commands/branch/create.test.ts
@@ -203,6 +203,50 @@ describe('branch create', () => {
     expect(() => JSON.parse(logs.join('\n'))).not.toThrow();
   });
 
+  it('switch failure after a successful create reports "switch failed", not "creation failed"', async () => {
+    const { getProjectConfig } = await import('../../lib/config.js');
+    (getProjectConfig as any).mockReturnValue({
+      project_id: 'p1',
+      project_name: 'parent',
+      org_id: 'o1',
+      appkey: 'p1ky',
+      region: 'us-east',
+      api_key: 'k',
+      oss_host: 'https://p1ky.us-east.insforge.app',
+    });
+    const { runBranchSwitch } = await import('./switch.js');
+    (runBranchSwitch as any).mockRejectedValueOnce(new Error('network down'));
+
+    const program = new Command().exitOverride();
+    program.option('--json').option('--api-url <url>').option('-y, --yes');
+    registerBranchCreateCommand(program);
+    let exitCode: number | undefined;
+    const origExit = process.exit;
+    (process.exit as any) = (code?: number) => {
+      exitCode = code;
+      throw new Error('__exit__');
+    };
+    const origStderr = process.stderr.write.bind(process.stderr);
+    process.stderr.write = (() => true) as any;
+    try {
+      await program
+        .parseAsync(['create', 'feat-x', '--mode', 'full'], { from: 'user' })
+        .catch(() => {});
+    } finally {
+      process.exit = origExit;
+      process.stderr.write = origStderr;
+    }
+    expect(exitCode).toBe(1);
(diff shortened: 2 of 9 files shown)


jwfing (commented):
## Code Review — InsForge/CLI#112

**Summary:** Solid bug fix and UX improvement. The `https://` scheme fix in `switch.ts` is correct and the extraction of `buildOssHost` into `config.ts` properly eliminates three independent copies of the same string-building logic. The clack spinner in `branch create` is a clear UX win over the previous silent hang.

---

### Requirements context

No `/docs/superpowers/` directory exists in this repo. Review assessed against the PR description and code intent directly.

---

### Findings

#### Critical

*(none)*

---

#### Suggestion

**[Software Engineering] `buildOssHost` has no direct unit test** — `src/lib/config.ts:110`

The bug being fixed (missing `https://` scheme) was caused by an unchecked string expression. `buildOssHost` is now the canonical implementation used by four call sites, but it has no unit test of its own. The switch test (`switch.test.ts:11`) exercises the behavior end-to-end, but it does so via a mock that independently hard-codes the expected format — it doesn't call the real function. A trivial test in a `config.test.ts` would give early warning if the function's output ever drifts:

```ts
it('buildOssHost always returns an https URL', () => {
  expect(buildOssHost('p1ky-x9p', 'us-east')).toBe('https://p1ky-x9p.us-east.insforge.app');
});
```

---

**[Software Engineering] No test covers the interactive (non-JSON) spinner path** — `src/commands/branch/create.ts:38-64`

All three tests in `create.test.ts` invoke the command with `--json`, which sets `spinner = null` and skips every `clack` call. The spinner refactoring — `spinner.start`, `spinner.message` inside `pollUntilReady`, `spinner.stop` on success and error — is exercised zero times by the test suite. If `@clack/prompts` is not mocked and a future test removes `--json`, the spinner will hit a real TTY and may hang or throw. Consider adding one non-JSON happy-path test with `@clack/prompts` mocked (the mock used in `link.test.ts` can serve as a template):

```ts
vi.mock('@clack/prompts', () => ({
  spinner: () => ({ start: vi.fn(), message: vi.fn(), stop: vi.fn() }),
}));
```

---

#### Information

- **No security-relevant changes.** All external inputs (branch name, mode) were already validated upstream; no new user input reaches SQL, shell, or HTTP without validation. `@clack/prompts` is an existing dependency (not newly introduced).

- **No performance concerns.** `POLL_INTERVAL_MS` and `POLL_TIMEOUT_MS` are unchanged; the spinner merely replaces `outputInfo` calls inside the polling loop.

- **`src/lib/config.ts:103-111` JSDoc is well-written.** The comment explains *why* the scheme is required (`fetch()` rejects bare hostnames), which is exactly the right level of documentation for a bug-prevention helper.

- **Version bump (0.1.69 → 0.1.70) in `package.json`.** Consistent with what appears to be the project's release-in-PR pattern; nothing wrong here, just noting it is included.

---

### Verdict

**approved** *(informational — explicit GitHub approval is a separate human action)*

No Critical findings. The two Suggestions (direct unit test for `buildOssHost`, non-JSON test for the spinner path) are worth addressing in a follow-up but are not blocking.

Fermionic-Lyu (approved):

PR #108, opened 2026-05-07
fix(db): unmask password when fetching database connection string

tonychang04:
## Summary

The platform endpoint \`/api/metadata/database-connection-string\` returns the URL with the password **masked as \`********\`** for safety. The CLI was passing this masked URL through unchanged:

- \`insforge db connection-string\` printed it (unusable for \`psql\`)
- \`link --auth better-auth\`'s auto-fill (0.1.67) wrote it to \`.env.local\` (BA's pg pool can't authenticate with \`********\`)

Hence after the previous round of fixes landed, users still hit \"can't connect\" — even when the URL was technically being filled.

## Fix

\`getDatabaseConnectionString()\` now also calls \`/api/metadata/database-password\` in parallel and splices the unmasked password into the URL between \`://<user>:\` and \`@\`. Self-hosted projects still get null (both endpoints 500 on missing PROJECT_ID), preserving the localhost default.

## Test plan

- [x] build + lint pass
- [ ] (post-merge, on cloud project) \`insforge db connection-string\` prints a working URL → \`psql \"$URL\" -c 'SELECT 1'\` returns 1
- [ ] (post-merge, on cloud project) \`link --auth better-auth\` writes a working URL → \`npm run setup\` runs \`auth:migrate\` against the cloud DB without auth failure

Bump 0.1.68 → 0.1.69.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## Summary by cubic
Unmasks the cloud database password so `insforge db connection-string` and `.env.local` auto-fill produce a working Postgres URL, and auto-runs `npm run setup` after `link --auth` to create tables. Self-hosted behavior is unchanged.

- **Bug Fixes**
  - `getDatabaseConnectionString()` now fetches `/api/metadata/database-connection-string` and `/api/metadata/database-password` in parallel, then splices the real password; `insforge db connection-string` prints/returns the unmasked URL, shows a clear error on self-hosted projects, and JSON output returns `{ connectionURL }`; adds unit tests for `spliceDatabasePassword`, `extractEnvPairs`, and `refreshStaleEnvDefaults`.
  - After `link --auth <provider>`, runs `npm run setup` if the scaffold added it to apply schema and migrations; failures are surfaced with guidance but don’t stop the link flow.

<sup>Written for commit a67a1ecb4dabcf4efffbfffbcd780917f153c061. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
[...]

PR #117, opened 2026-05-12
fix(link): refresh masked-password URLs on re-link

tonychang04:
## Why

Pre-0.1.72 cloud \`link --auth\` wrote DATABASE_URL with a literal placeholder password:

\`\`\`
DATABASE_URL=postgresql://postgres:********@m8s5kmam.us-east.database.insforge.app:5432/insforge?sslmode=require
\`\`\`

The cloud's \`/database-connection-string\` endpoint masks the password as \`********\`; the unmask via the separate \`/database-password\` endpoint only landed in #108 (shipped in 0.1.72).

Users who linked under an older CLI still have the masked URL in \`.env.local\`. Re-linking under 0.1.72+ doesn't fix it because \`refreshStaleEnvDefaults\` only replaces values that match the manifest's literal default (\`postgresql://postgres:postgres@127.0.0.1:5432/insforge\`). The masked-cloud URL matches neither the default nor a real user customization — it falls into a gap.

The symptom is BA's pg pool failing on every sign-up:

\`\`\`
password authentication failed for user "postgres"
\`\`\`

## What this PR does

Treat any \`:****@\` (one or more \`*\`) password pattern as stale and refresh from the platform value, alongside the existing "matches the manifest default" branch.

\`\`\`diff
-    if (def !== undefined && real !== undefined && userValue === def && real !== def) {
+    if (def === undefined || real === undefined || real === def) return line;
+    const matchesDefault = userValue === def;
+    const isMaskedPassword = MASKED_PASSWORD_PATTERN.test(userValue);
+    if (matchesDefault || isMaskedPassword) {
       refreshed.push(key);
       return \`\${key}=\${real}\`;
     }
\`\`\`

The masked password is never a valid Postgres credential, so the "treat as stale" carries no risk of clobbering real user data.

## Test plan
- [x] Unit test added: \`refreshes a masked-password URL even if it differs from the manifest default\`
- [x] 14/14 \`apply.test.ts\` tests pass
- [x] Existing "preserves user value when it differs from the manifest default" test still passes — a real user customization without the \`:****@\` pattern is left alone
- [ ] Manual: re-link a project that has the masked URL → \`.env.local\` gets the real password spliced in → \`npm run setup\` succeeds

---
## Summary by cubic
Fixes re-linking so masked `DATABASE_URL` values from older cloud links are replaced with the real password. Detects `:****@` patterns and refreshes from platform values to prevent Postgres auth failures.

- **Bug Fixes**
  - Extend `refreshStaleEnvDefaults` to refresh values with masked passwords (`:\*+@`) in addition to default-match cases.
  - Preserve real user edits and add a unit test for the masked-URL refresh case.

<sup>Written for commit 50709e252d3014ca92ac92276473341735ead5f2. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed refresh handling for database connection strings containing masked passwords in environment configuration files. Previously, connection strings with placeholder passwords were not being updated to use actual platform credentials.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/CLI/pull/117)

### DIFF
--- a/src/auth-providers/apply.test.ts
+++ b/src/auth-providers/apply.test.ts
@@ -100,6 +100,23 @@ describe('refreshStaleEnvDefaults', () => {
     expect(updated).toBe(existing);
   });
 
+  it('refreshes a masked-password URL even if it differs from the manifest default', () => {
+    // Pre-0.1.72 cloud links wrote `postgresql://postgres:********@...`
+    // (placeholder password) to .env.local. Re-linking under 0.1.72+ should
+    // replace that with the real spliced password from /database-password.
+    const existing = 'DATABASE_URL=postgresql://postgres:********@m8s5kmam.us-east.database.insforge.app:5432/insforge?sslmode=require\n';
+    const defaults = new Map([
+      ['DATABASE_URL', 'postgresql://postgres:postgres@127.0.0.1:5432/insforge'],
+    ]);
+    const platform = new Map([
+      ['DATABASE_URL', 'postgresql://postgres:realsecret@m8s5kmam.us-east.database.insforge.app:5432/insforge?sslmode=require'],
+    ]);
+    const { updated, refreshed } = refreshStaleEnvDefaults(existing, defaults, platform);
+    expect(refreshed).toEqual(['DATABASE_URL']);
+    expect(updated).toContain('postgres:realsecret@');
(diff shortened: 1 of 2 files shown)


jwfing (approved):
LGTM, approved.

PR #120, opened 2026-05-12
feat(link): install skills only when called with no args

CarmenDou:
---
## Summary by cubic
Running `projects link` with no args now installs InsForge agent skills only, skipping auth and project linking. This creates a fast, sign-in-free quick start; the agent will guide provisioning later.

- **New Features**
  - Bare `link` installs skills in the current directory and exits; with `--json` it returns `{ success: true, skills_only: true }`.
  - Skips auth, org/project picker, and writing `.insforge/project.json`; providing flags like `--project-id` bypasses this path.
  - Updates help text and shows a short next-steps note in the CLI.
  - Sends analytics (`trackCommand`, `reportCliUsage`) and exits with code 1 on install errors; detection uses explicit per-option checks with tests for no-args, `--json`, error, and bypass cases.

<sup>Written for commit 1a0648cb7ebf4d01c3948d5a871a8b0b0bb6c639. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **Documentation**
  * Updated command description to clarify behavior when called without arguments.

* **New Features**
  * Added optimized execution path for no-argument invocation, streamlining the agent skills installation workflow with enhanced error handling and reporting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/CLI/pull/120)

### DIFF
--- a/src/commands/projects/link.test.ts
+++ b/src/commands/projects/link.test.ts
@@ -1,0 +1,163 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { Command } from 'commander';
+import { registerProjectLinkCommand } from './link.js';
+
+vi.mock('../../lib/skills.js', () => ({
+  installSkills: vi.fn(async () => {}),
+  reportCliUsage: vi.fn(async () => {}),
+}));
+
+vi.mock('../../lib/analytics.js', () => ({
+  captureEvent: vi.fn(),
+  trackCommand: vi.fn(),
+  shutdownAnalytics: vi.fn(async () => {}),
+}));
+
+vi.mock('../../lib/api/platform.js', () => ({
+  listOrganizations: vi.fn(),
+  listProjects: vi.fn(),
+  getProject: vi.fn(),
+  getProjectApiKey: vi.fn(),
+  reportAgentConnected: vi.fn(),
+}));
+
+vi.mock('../../lib/credentials.js', () => ({
+  requireAuth: vi.fn(),
+}));
+
+vi.mock('../../lib/config.js', () => ({
+  getGlobalConfig: vi.fn(() => ({})),
+  saveGlobalConfig: vi.fn(),
+  saveProjectConfig: vi.fn(),
+  getFrontendUrl: () => 'https://example.test',
+  buildOssHost: vi.fn(),
+  FAKE_PROJECT_ID: '00000000-0000-0000-0000-000000000000',
+  FAKE_ORG_ID: '00000000-0000-0000-0000-000000000000',
+}));
+
+vi.mock('../../lib/output.js', () => ({
+  outputJson: vi.fn(),
+  outputSuccess: vi.fn(),
+}));
+
+vi.mock('../create.js', () => ({
+  downloadGitHubTemplate: vi.fn(),
+}));
+
+vi.mock('../../auth-providers/apply.js', () => ({
+  applyAuthProvider: vi.fn(),
+  VALID_AUTH_PROVIDERS: ['better-auth'],
+}));
+
+vi.mock('../../lib/prompts.js', () => ({
+  text: vi.fn(),
+  select: vi.fn(),
+  isCancel: () => false,
+}));
+
+vi.mock('@clack/prompts', () => ({
+  spinner: () => ({ start: vi.fn(), message: vi.fn(), stop: vi.fn() }),
+  log: {
+    info: vi.fn(),
+    success: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
+    step: vi.fn(),
+  },
+  note: vi.fn(),
+}));
+
+function buildProgram(): Command {
+  const program = new Command().exitOverride();
+  // Mirror the global flags the action handler reads via getRootOpts(cmd).
+  program.option('--json').option('--api-url <url>');
+  registerProjectLinkCommand(program);
+  return program;
+}
+
+describe('project link: skills-only fast path', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('with no args, installs skills and skips auth + project picker', async () => {
+    const program = buildProgram();
+    await program.parseAsync(['link'], { from: 'user' });
+
+    const { installSkills, reportCliUsage } = await import('../../lib/skills.js');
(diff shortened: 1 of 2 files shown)


jwfing (changes requested):
## Summary

Adds a clean "skills-only fast path" to `projects link` — when called with no args it installs agent skills without requiring auth, org/project picker, or writing `.insforge/project.json`. The intent is clear and the implementation is mostly correct, but it ships without any tests.

---

## Requirements context

No `/docs/superpowers/` directory exists in this repo. Review assessed against the PR description alone, cross-referenced with the project's TDD convention (insforge-development skill) which requires **every new behavior to have at least one test**.

---

## Findings

### Critical

**[Software Engineering] No test coverage for the new fast path**
`src/commands/projects/link.ts` (lines 103–126)

There is no `link.test.ts` file anywhere in the repo (confirmed via glob). The project has test files for `create`, all `branch/*` subcommands, and various `lib/` modules — the `link` command is the only command-level entrypoint with zero tests. This PR adds an entirely new execution branch (skills-only mode) with no tests covering:
- The happy path (no args → `installSkills` called, correct output)
- The `--json` flag variant (no args + `--json` → `{ success: true, skills_only: true }`)
- The error path (`installSkills` throws → `reportCliUsage(false)` + `handleError` called)

Per the project's TDD convention: *"Every new behavior has at least one test."* Please add a `link.test.ts` (or equivalent) before merging.

---

### Suggestion

**[Functionality] `isSkillsOnly` detection is fragile**
`src/commands/projects/link.ts:103`

```ts
const isSkillsOnly = Object.values(opts).every((v) => v === undefined);
```

This works correctly today because every declared option uses a `<value>` placeholder and defaults to `undefined`. But it will silently break if a future contributor adds:
- a boolean flag with no placeholder (e.g. `.option('--verbose')` → defaults to `false`, not `undefined`)
- an option with an explicit default value

A more robust form makes the intent explicit and is immune to future option additions:

```ts
const isSkillsOnly =
  opts.projectId === undefined &&
  opts.orgId === undefined &&
  opts.template === undefined &&
  opts.auth === undefined &&
  opts.apiBaseUrl === undefined &&
  opts.apiKey === undefined;
```

---

### Information

**[Software Engineering] Double `shutdownAnalytics()` in the skills-only error path**
`src/commands/projects/link.ts:122` and `src/commands/projects/link.ts:537`

When `installSkills` throws, `shutdownAnalytics()` is called explicitly at line 122, then again in the outer `finally` at line 537. This is consistent with the existing pattern for the direct-link error path (lines 300–303) so it's not a regression, but it's a pre-existing code smell worth tracking.

---

## Verdict

`request_changes` — one Critical finding (missing tests for new behavior).

jwfing (commented):
## Code Review — `feat(link): install skills only when called with no args`

**Summary:** The PR adds a clean "skills-only fast path" to `projects link` that skips auth, org/project selection, and config writing when the command is invoked with no options, installing agent skills immediately instead.

**Requirements context:** No `/docs/superpowers/` directory exists in this repo. Review assessed against the PR title, body, and the implementation itself.

---

### Findings

#### Critical

*(none)*

---

#### Suggestion

**Software engineering — `isSkillsOnly` is a growing enumeration that can silently exclude future flags**
`src/commands/projects/link.ts:103–109`

The fast-path gate enumerates every option explicitly:

```ts
const isSkillsOnly =
  opts.projectId === undefined &&
  opts.orgId === undefined &&
  opts.template === undefined &&
  opts.auth === undefined &&
  opts.apiBaseUrl === undefined &&
  opts.apiKey === undefined;
```

This is correct today, but any new option added to `link` in the future won't automatically be excluded from the fast path — the maintainer must remember to extend this list. A short comment (`// extend this check when adding new options`) would prevent a silent regression. Alternatively, you could guard with an assertion in the option definitions themselves, but the comment is the lower-friction fix.

---

**Software engineering — `process.exit` monkey-patching in tests**
`src/commands/projects/link.test.ts:112–128, 141–153`

The error and bypass tests monkey-patch `process.exit` by direct assignment:

```ts
(process.exit as unknown) = (code?: number) => { … throw new Error('__exit__'); };
```

This pattern works, but vitest provides `vi.spyOn(process, 'exit').mockImplementation(…)` which is more idiomatic and automatically restores on `vi.clearAllMocks()` / `vi.restoreAllMocks()` in the `beforeEach`. The current `try/finally` restore is correct, but `vi.spyOn` would be safer if an earlier line inside `try` throws before reaching the mutation.

---

**Software engineering — dynamic module imports inside test bodies**
`src/commands/projects/link.test.ts:88–93, 103–104, 118, 134–135, 148–149`

After `parseAsync`, each test does:

```ts
const { installSkills } = await import('../../lib/skills.js');
```

Since `vi.mock` is hoisted, the mock reference is stable and can be imported once at the top of the file with a `const { installSkills } = await import(…)` or by capturing the return value of `vi.mock`. The repeated dynamic imports add noise without benefit and are mildly confusing to future readers.

---

#### Information

**Functionality — `return` after `handleError` in the inner catch is dead code**
`src/commands/projects/link.ts:129–131`

```ts
handleError(err, json);
return;
```

`handleError` calls `process.exit(1)`, so the `return` is never reached. Harmless, and consistent with the existing `link_direct` error path, so no action required — just noting it.

---

**Functionality — `trackCommand` is not awaited**
`src/commands/projects/link.ts:114`

```ts
trackCommand('link', 'skills-only', { skills_only: true });
```

Not awaited, matching the pattern in all other link paths (lines 291, 402). Consistent with the existing convention; fine as-is.

---

**Functionality — outer `finally` handles `shutdownAnalytics` on success**

The skills-only success path calls `return` (line 125), which exits the inner `try` and still triggers the outer `finally` at line 543 — so `shutdownAnalytics()` is correctly called on both success and error without duplication on the success path. The error path does call it twice (inner catch + outer finally), but `shutdownAnalytics` is idempotent. No action needed.

---

### Verdict

**Approved** *(informational — no Critical findings; human sign-off still required via the approve flow).*

The fast path logic is correct, the `isSkillsOnly` gate is explicit and tested, and the 4 tests cover the key behaviors (no-args, `--json`, error/exit-1, bypass via `--project-id`). The suggestions above are polish-level; none block merging.

jwfing (approved):
LGTM, approved.

PR #109, opened 2026-05-07
feat(cli): config-as-code (export/plan/apply) + env() validation + compute port fix

tonychang04:
**Companion to:** `InsForge/InsForge#1216` (backend half — extends \`/api/metadata\` to surface \`auth.allowed_redirect_urls\`)

## What

Adds three CLI commands that round-trip a single TOML field — `[auth] allowed_redirect_urls` — between \`insforge.toml\` and the live project state:

\`\`\`bash
insforge config export    # GET /api/metadata → write insforge.toml
insforge config plan      # diff TOML vs metadata, print plan
insforge config apply     # plan → confirm → PUT /api/auth/config
\`\`\`

This is the smallest possible vertical slice of the config-as-code feature. Everything else (SMTP, OAuth providers, custom domains, etc.) is matrix-fill on top of this architecture.

## Why

Agents can't make fine-grained config changes today — setting up a custom domain, configuring SMTP, changing an auth redirect URL all require dashboard clicks (invisible, not reproducible) or per-field CLI subcommands (sprawls fast, no PR diff). With ~100 config knobs across the platform, per-field CLI is the wrong shape.

This PR ships the TOML + apply loop with the smallest possible scope so the architecture proves out before we widen the surface. Full design in \`spec-toml-mvp.md\` and \`proposal-insforge-config-toml.md\`.

## What this PR contains

9 commits, all on \`feat/insforge-toml-mvp\` off \`main\`:

- \`chore(cli): add smol-toml for insforge.toml parsing\`
- \`feat(cli): add config schema for v1 MVP\` — types + validator
- \`feat(cli): add TOML <-> JSON codec\` — parse + stringify with stable ordering
- \`feat(cli): add structured per-section config diff\` — default-keep semantics
- \`feat(cli): add human-readable plan formatter\` — human + JSON output
- \`feat(cli): scaffold insforge config subcommand\` — registrar in src/index.ts
- \`feat(cli): add insforge config export\`
- \`feat(cli): add insforge config plan\`
- \`feat(cli): add insforge config apply\` — plan-confirm-apply with --dry-run / --auto-approve

**Lib:** 6 files in \`src/lib/config-{schema,toml,diff,format}.ts\` + tests
**Commands:** 4 files in \`src/commands/config/{index,export,plan,apply}.ts\`
**Tests:** 12 vitest tests, all green

## What this PR does NOT do

- No new server endpoints (read uses existing \`/api/metadata\`; write uses existing \`PUT /api/auth/config\`)
[...]

PR #123, opened 2026-05-12
feat(cli): add [auth.smtp] and [deployments] sections to config-as-code

tonychang04:
## Summary

Step 2 of the SMTP config-as-code initiative. Adds `[auth.smtp]` to the declarative TOML surface with secret-aware password handling.

Companion PR: **InsForge/InsForge#1258** (backend — exposes `smtpConfig` in admin `/api/metadata`). Required for this CLI feature to work — without it, `config apply` correctly puts SMTP changes in \`skipped[]\` per the capability gate.

### Schema
- New `[auth.smtp]` section in `insforge.toml` with 8 fields (enabled, host, port, username, password, sender_email, sender_name, min_interval_seconds)
- `password` field validated as `env(NAME)` ref only — literal values rejected at parse time (the validator infrastructure from PR #109 finally gets its first consumer)

### Capability gate
- `metadataSupports()` extended to probe `raw.auth.smtpConfig` presence
- SMTP changes against an older backend (no smtpConfig in metadata) land in `skipped[]` with the upgrade message; **no PUT issued** — same protocol as redirect URLs

### env() ref resolution
- New `resolveEnvRef(envRef, fieldPath)` in `src/lib/config-secrets.ts`
- Calls `/api/secrets/:name` pre-flight before any PUT — missing/inactive secrets fail fast with actionable error (`SECRET_NOT_FOUND` / `SECRET_EMPTY`) naming the exact `secrets add` command
- **Resolution source is InsForge secrets store only** (not local env vars) — secrets are shared per-project across teammates and CI

### Diff layer
- `DiffChange` discriminated union: `auth` (redirect URLs) | `auth.smtp` (whole-object upsert)
- **Force-resend password semantics**: when TOML carries `password = \"env(NAME)\"`, every \`apply\` resolves the current secrets-store value and re-sends. Reason: backend ciphertext can't be diffed against client-side, and re-sending matches user mental model (\"I just rotated SMTP_PASSWORD, re-apply should propagate it\")
- Password slot in plan output is always an opaque marker (\`(set)\`, \`(unset)\`, \`env(NAME)\`, \`(unchanged)\`) — actual values never appear

### Export
- When backend reports \`hasPassword: true\`, emits \`password = \"env(SMTP_PASSWORD)\"\` as a deterministic placeholder + discovery comment (\`# password is managed via secrets — run \\\`insforge secrets add SMTP_PASSWORD <value>\\\`\`) so the workflow is learnable from a fresh export without reading skill docs
- When \`hasPassword: false\`, omits the password line entirely

### Apply
- New variant in \`applyChange()\` for \`auth.smtp\`: builds the upsert body, resolves env() ref pre-flight, PUTs \`/api/auth/smtp-config\`
- Omitting \`password\` from the TOML omits it from the PUT body — backend's upsert preserves the existing encrypted value (matches our \"absent = preserve\" semantics)

## Version bump

0.1.75 → 0.1.76

## Tests

- **51 config-flow unit tests pass** (was 38 pre-SMTP)
- New coverage: SMTP schema parsing + literal-password rejection, TOML roundtrip with placeholder, diff force-resend semantics, env() resolution failure modes (404 / empty / non-200), apply path with secrets-store mock, capability skip when backend predates the field
- Build clean (\`npm run build\`)
- Lint clean (\`npx eslint src/\` — 0 errors; 39 pre-existing warnings unrelated)

## Test plan

- [x] Unit tests cover schema, capability gate, diff, env() resolution, apply, export
- [ ] E2E against local backend with #1258 (deferred — docker not available locally this session; recommend post-merge of #1258, before tagging the npm release)
- [ ] Run \`config apply\` against a cloud project once #1258 is deployed

## Design notes captured in code

The trickier decisions are documented inline:
- Force-resend rationale: \`src/lib/config-diff.ts\` \`diffSmtp()\`
- env() resolution from InsForge secrets store (not local env): \`src/lib/config-secrets.ts\` \`resolveEnvRef()\` JSDoc
- Default-keep vs explicit-set for password: \`src/commands/config/apply.ts\` \`applyChange()\` comment
- Why password is never round-tripped on export: \`src/commands/config/export.ts\` SMTP block

## Followups (out of scope here)

- **Skills PR**: re-introduce the SMTP TOML example we trimmed last session — now it actually works. Document the placeholder convention, \`hasPassword\` round-trip, and force-resend plan output.
- **E2E** after #1258 is deployed.

---
## Summary by cubic
Adds `[auth.smtp]` and `[deployments]` subdomain to `insforge.toml` with secure `env(NAME)` password handling, export/apply wiring, and capability gates. Also tightens validation and improves export hints.

- **New Features**
  - Schema: `[auth.smtp]` with `enabled`, `host`, `port` (1–65535), `username`, `password` (must be `env(NAME)`), `sender_email`, `sender_name`, `min_interval_seconds`; `[deployments]` with `subdomain` (string or null; `""` clears on apply).
  - Apply: SMTP → PUT `/api/auth/smtp-config` (omit `password` to preserve; `password = "env(NAME)"` resolves via secrets store and force-resends); Deployments → PUT `/api/deployments/slug` with `{ slug: "name" | null }`.
  - Export: SMTP emits `password = "env(SMTP_PASSWORD)"` when a password exists, with a discovery comment that names the actual secret in the file; Deployments emits only when a slug is set.
  - Capability gate: applies/exports only when `/api/metadata` exposes `auth.smtpConfig` and `deployments`; otherwise entries are added to `skipped[]`.

- **Bug Fixes**
  - Secret lookup recovers missing-secret signals from `ossFetch`-thrown errors so missing secrets surface as `SECRET_NOT_FOUND` with an `insforge secrets add NAME "<value>"` hint.
  - Export typeof-guards `auth` before checking `smtpConfig` to avoid a crash on malformed metadata.
  - Capability checks add an exhaustiveness guard in `metadataSupports` to prevent silently skipping future sections.

<sup>Written for commit 6d6eee4d6be20aaa2aed2c282bec0a673fb85eec. Summary will update on new commits.</sup>

## Summary by CodeRabbit

* **New Features**
  * Add SMTP support in config: export and apply flows, SMTP-specific diffs/formatting, and deployments subdomain handling; password fields use env-secret references and are preserved when omitted.
* **Tests**
  * Expanded coverage for SMTP behavior, secret resolution, exports, apply flows, and deployments subdomain logic.
* **Chores**
  * Bumped package version to 0.1.77.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/CLI/pull/123)

> [!NOTE]
> ### Add [auth.smtp] and [deployments] sections to config-as-code export and apply
> - Extends the TOML schema, diff, export, and apply pipeline to support two new config sections: `[auth.smtp]` and `[deployments]`.
> - Export writes `[auth.smtp]` fields with an `env(SMTP_PASSWORD)` placeholder when a password exists server-side, and writes `[deployments]` when a non-empty custom subdomain is set.
> - Apply resolves `env()` secret references via `GET /api/secrets/:name` before issuing `PUT /api/auth/smtp-config`, failing fast with actionable errors if the secret is missing or empty. Deployment subdomain changes are sent via `PUT /api/deployments/slug`, including `null` to clear.
> - Both sections are skipped with a named reason when the backend metadata does not expose the corresponding capability.
> - Behavioral Change: literal passwords in TOML are now rejected at parse time; only `env()` references are accepted for the SMTP password field.
>
> 
> #### Changes since #123 opened
>
> - Tightened SMTP port validation to require values within the 1-65535 range [6d6eee4]
> - Modified SMTP password comment generation to use actual environment variable reference name [6d6eee4]
> - Added exhaustiveness check for DiffChange variants in config-capabilities [6d6eee4]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 35af5b9.</sup>
> 
>
>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@insforge/cli",
-  "version": "0.1.75",
+  "version": "0.1.76",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@insforge/cli",
-      "version": "0.1.75",
+      "version": "0.1.76",
       "license": "Apache-2.0",
       "dependencies": {
         "@clack/prompts": "^0.9.1",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@insforge/cli",
-  "version": "0.1.75",
+  "version": "0.1.76",
   "description": "InsForge CLI - Command line tool for InsForge platform",
   "type": "module",
   "bin": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@insforge/cli",
-  "version": "0.1.76",
+  "version": "0.1.77",
   "description": "InsForge CLI - Command line tool for InsForge platform",
   "type": "module",
   "bin": {
--- a/src/commands/config/apply.test.ts
+++ b/src/commands/config/apply.test.ts
@@ -8,9 +8,23 @@ import type * as ErrorsModule from '../../lib/errors.js';
 
 // Per-test we override what /api/metadata returns by reassigning this.
 let nextMetadataResponse: unknown = {};
+// Stash secret values for env() resolution. Map secret name → value.
+const secretsStore: Map<string, string> = new Map();
 const ossFetchMock = vi.fn(async (path: string, init?: RequestInit) => {
   if (path === '/api/metadata' && (!init || init.method === undefined || init.method === 'GET')) {
     return new Response(JSON.stringify(nextMetadataResponse), {
+      status: 200,
+      headers: { 'content-type': 'application/json' },
+    });
+  }
+  const secretMatch = path.match(/^\/api\/secrets\/(.+)$/);
+  if (secretMatch && (!init || init.method === undefined || init.method === 'GET')) {
+    const key = decodeURIComponent(secretMatch[1]);
+    const value = secretsStore.get(key);
+    if (value === undefined) {
+      return new Response(JSON.stringify({ error: 'not found' }), { status: 404 });
+    }
+    return new Response(JSON.stringify({ key, value }), {
       status: 200,
       headers: { 'content-type': 'application/json' },
     });
@@ -75,6 +89,7 @@ let tmp: string;
 
 beforeEach(() => {
   vi.clearAllMocks();
+  secretsStore.clear();
   tmp = mkdtempSync(join(tmpdir(), 'insforge-apply-test-'));
 });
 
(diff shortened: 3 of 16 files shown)


jwfing (commented):
## Code Review — feat(cli): add [auth.smtp] section to config-as-code with env() password resolution

**Summary:** Solid, well-scoped addition of SMTP config-as-code. The security model (env()-only passwords, opaque plan output, pre-flight secret resolution, no plaintext in logs) is correct and the test coverage is thorough. Four non-blocking items below.

**Requirements context:** No `/docs/superpowers/` directory exists in this repo — assessing against the PR description (Step 2 of SMTP config-as-code initiative) and companion PR InsForge/InsForge#1258 as the stated requirements. The implementation matches the spec described in the PR body on all major points: schema, capability gate, env() resolution, diff/force-resend semantics, export placeholder, and apply upsert path.

---

## Findings

### Critical

_(none)_

---

### Suggestion

**[Software Engineering / Functionality] Port range not validated — `src/lib/config-schema.ts:113–118`**

`validateSmtp` accepts any integer for `port` (including 0, negatives, or values > 65535). Compare the existing non-negative guard on `min_interval_seconds` at line 151–162 — the same pattern should apply here:

```ts
if (
  typeof obj.port !== 'number' ||
  !Number.isInteger(obj.port) ||
  obj.port < 1 ||
  obj.port > 65535
) {
  throw new ConfigValidationError('auth.smtp.port', 'must be an integer between 1 and 65535');
}
```

A user who writes `port = 0` will get a cryptic backend 400 rather than a parse-time error. Low blast radius but inconsistent with how `min_interval_seconds` is treated in the same function.

---

**[Functionality] Discovery comment in TOML renderer hardcodes `SMTP_PASSWORD` regardless of the actual ref — `src/lib/config-toml.ts:60`**

```ts
lines.push('# password is managed via secrets — run `insforge secrets add SMTP_PASSWORD <value>`');
lines.push(`password = ${JSON.stringify(smtp.password)}`);
```

`stringifyConfigToml` is a general renderer, but the comment always mentions `SMTP_PASSWORD`. If `smtp.password` is `"env(PROD_SMTP_PASS)"`, the output becomes:

```toml
# password is managed via secrets — run `insforge secrets add SMTP_PASSWORD <value>`
password = "env(PROD_SMTP_PASS)"
```

The comment is misleading — it tells the user to create a secret under the wrong name. Extract the name from the env() ref:

```ts
const ref = parseEnvRef(smtp.password);
const secretName = ref ?? 'SMTP_PASSWORD';
lines.push(`# password is managed via secrets — run \`insforge secrets add ${secretName} "<value>"\``);
```

_(Export-generated TOML always uses `env(SMTP_PASSWORD)` today so the comment is correct there — but this will bite the first user who names their secret differently.)_

---

**[Functionality] Secret-not-found detection relies on error message string — `src/lib/config-secrets.ts:108–116`**

```ts
const message = (err as Error).message ?? '';
if (/not found/i.test(message)) {
```

This creates implicit coupling to `ossFetch`'s error message format. If the backend error wording changes (e.g. "Secret 'X' does not exist"), the regex fails to match and the error silently degrades from the actionable `SECRET_NOT_FOUND` (with `insforge secrets add` hint) to the generic `SECRET_LOOKUP_FAILED`. The mock in the test (`new Error('Secret not found: MISSING')`) hard-codes the matching string, so a backend message change wouldn't be caught.

Not urgent since it only affects error quality, not correctness — but it's worth tracking. Ideally `ossFetch` would surface the HTTP status so callers can branch on 404 vs 5xx without string-matching.

---

**[Software Engineering] `metadataSupports` has no exhaustiveness check on the DiffChange union — `src/lib/config-capabilities.ts:56`**

Unlike `applyChange()` (which has the `const _exhaustive: never = change` guard at line 205–206 of `apply.ts`), `metadataSupports` silently returns `false` for any unrecognized `change.section`. If a new DiffChange variant is added in a future PR, `metadataSupports` will silently skip it (every apply of that new section will be put into `skipped[]`). Adding the same exhaustiveness pattern here would catch that at compile time.

---

### Information

**Duplicated `RawAuthMetadata` / `RawMetadataResponse` interfaces — `apply.ts:17–29` and `export.ts:14–26`**

Identical interface definitions in two files. Not worth splitting into a shared module for two consumers, but flag for the next person who adds a third field — that's the right moment to consolidate.

---

## Verdict

**Approved** _(informational — explicit GitHub approval is a separate human action)_

No Critical findings. Four non-blocking items. The secret-handling model is well-designed, test coverage is strong, and scope discipline is clean. The port-range validation and comment-rendering issues are the most concrete and worth a quick follow-up before the npm release (0.1.76 tag).

jwfing (commented):
## Code Review — `feat(cli): add [auth.smtp] and [deployments] sections to config-as-code`

**Summary:** Well-structured step-2 implementation of the SMTP config-as-code initiative. The capability-gate protocol, force-resend semantics, and secret-opaque plan output are all correct; test coverage is solid across the new surface area.

---

### Requirements context

No `/docs/superpowers/` directory exists in this repo — assessed against the PR description alone. The PR is self-contained with clear design rationale; all claimed behaviors were verified against the implementation.

---

### Findings

#### Critical

_(none)_

---

#### Suggestion

**[Software Engineering] Duplicate `RawAuthMetadata` / `RawMetadataResponse` types — `apply.ts:17–39` and `export.ts:14–36`**

Both files define identical `RawAuthMetadata` and `RawMetadataResponse` interfaces. They already diverge slightly in comments and will diverge further as the metadata shape grows (e.g. future `[auth.providers]` section). Extracting them once to `src/lib/config-metadata.ts` (or similar) prevents silent drift and makes both callers authoritative off the same type.

---

**[Functionality] Fragile "not-found" detection in `resolveEnvRef` — `config-secrets.ts:103–116`**

```ts
const message = (err as Error).message ?? '';
if (/not found/i.test(message)) {
```

`ossFetch` throws for _any_ non-2xx, swallowing the HTTP status. The regex `/not found/i` recovers the 404 case from the message string, but it will also match unrelated failures — for example, a reverse-proxy or network error whose message contains "route not found", "host not found", or a secret name that happens to contain "not found". The user then receives a misleading `insforge secrets add` hint when the actual problem is connectivity.

More robust: have `ossFetch` (or a wrapper) attach the numeric status to the thrown error so callers can branch on `err.status === 404` rather than string-matching. If that's out of scope here, the fallback fix is a tighter pattern — e.g. `/^secret not found:/i` to match only the backend's exact error format, and leave all other throw paths in the `SECRET_LOOKUP_FAILED` branch.

The test at `config-secrets.test.ts:100` deliberately mirrors the current throw format, which means this assumption is tested — but the test would also pass if the regex were accidentally over-broad.

---

**[Functionality] Omitting SMTP fields from TOML silently applies renderer defaults — `config-diff.ts:230–241` + `apply.ts:186–194`**

`renderFileSmtp` fills absent fields with hardcoded defaults (`port: 587`, `min_interval_seconds: 60`, `sender_email: ''`, …). Because `applyChange` builds the upsert body from `change.to` (the full rendered view), a minimal TOML like:

```toml
[auth.smtp]
enabled = true
host = "smtp.example.com"
username = "user"
password = "env(SMTP_PASSWORD)"
```

will PUT `port=587`, `sender_email=''`, etc., _overwriting_ any non-default values the backend currently holds for those fields. This is whole-object upsert semantics and the design rationale is noted in the PR, but the plan output ("~ smtp config: …") only shows fields that _change_ — so a user with `port=25` on the backend will see `port: 25 → 587` in their plan, but may not understand why. A one-line note in the formatted output (e.g., _"omitted fields will be reset to defaults"_) would reduce surprise.

---

#### Information

**Version bump mismatch in PR description:** The body says "0.1.75 → 0.1.76" but `package.json` shows `"0.1.76" → "0.1.77"`. The code is correct; the description text is stale.

**Test coupling to `ossFetch` throw format — `config-secrets.test.ts:100`:** The mock reproduces the exact string `"Secret not found: MISSING"` to drive the `SECRET_NOT_FOUND` branch. If `ossFetch` is ever refactored to produce a different message format, this test continues to pass (it controls both sides of the mock) while production silently falls through to `SECRET_LOOKUP_FAILED`. Worth a comment noting the coupling.

**`stringifyConfigToml` — no test for fully-empty config (`config-toml.ts:23–61`):** The TOML serializer's final `replace(/\n+$/, '\n')` is exercised by every test with content, but there's no test for `stringifyConfigToml({})` (no sections at all). Minor gap.

---

### Verdict

**approved** _(informational — no Critical findings; explicit GitHub approval is a separate human action)_

The three Suggestion items are clean-up / robustness improvements rather than blockers. The core capability-gate protocol, secret-opaque diff output, and force-resend semantics are all implemented correctly. No security issues found.

jwfing (approved):
LGTM, approved.

PR #129, opened 2026-05-15
feat(posthog): delegate SDK install to @posthog/wizard

CarmenDou:
## Summary by CodeRabbit

* **New Features**
  * PostHog setup now uses a two-phase flow: establishing a dashboard connection followed by the official PostHog wizard for SDK integration.

* **Chores**
  * Simplified PostHog setup command interface and removed built-in framework/package detection and template management in favor of delegating to the PostHog wizard.

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/CLI/pull/129)


jwfing (changes requested):
## Summary

Good scope reduction overall — delegating SDK wiring to `@posthog/wizard` removes a lot of fragile framework-specific code, and the two-step flow is clearly explained. However, two blocking issues need to be addressed before this merges.

## Requirements context

No matching spec/plan found under `docs/specs/` (the three existing specs are for the diagnose command and db-migrations). Assessment is based on the PR description and body alone.

---

## Findings

### Critical

#### Software Engineering — No tests for new behavior

`src/commands/posthog/setup.ts` is the core of this PR (70 lines added, replacing 429), yet no test file is added or updated for it. The deleted `framework-detect.test.ts` is correctly removed alongside its subject, but the new two-step flow has zero coverage:

- `ensureDashboardConnection` — two code paths (`already-connected` fast path with sanity-fetch vs. OAuth + poll), each testable by mocking `startPosthogCliFlow` / `fetchPosthogConnection` / `pollPosthogConnection`.
- Wizard error handling — the `wizardResult.error` branch and the `exitCode !== 0` branch.
- `--json` flag interactions.

The `insforge-development` skill (this project's own TDD convention) is explicit: *"Every new behavior has at least one test. Every bug fix has a regression test."* No tests = does not meet the merge bar.

#### Functionality — JSON mode is non-functional (`setup.ts:100-111`)

```typescript
const wizardResult = spawnSync('npx', ['-y', '@posthog/wizard@latest'], {
  stdio: opts.json ? 'pipe' : 'inherit',   // <-- problem
  env: process.env,
});
```

`@posthog/wizard` is an interactive tool: it prompts for PostHog project selection and OAuth. With `stdio: 'pipe'`, the wizard's stdin is a closed/empty pipe — any read blocks indefinitely or gets EOF immediately, causing the wizard to exit non-zero. That triggers the `exitCode !== 0` branch and a `CLIError`, so `outputJson({ success: true, ...result })` never runs. **`insforge posthog setup --json` will always fail.**

The options to fix this are:
1. Run the wizard unconditionally with `stdio: 'inherit'` (dropping `--json` support for this command, which is documented as a behavioral change anyway), or
2. Document that `--json` only covers the dashboard-connection step and skip the wizard in JSON mode (returning early after `ensureDashboardConnection`), or
3. Use async `spawn` and stream wizard output to `process.stdout`/`stderr` while still capturing the exit code.

---

### Suggestion

#### Performance — `spawnSync` blocks the event loop for the full wizard duration (`setup.ts:100`)

`spawnSync` is synchronous and holds the V8 event loop until the wizard exits — which involves OAuth round-trips and package installation, potentially 5+ minutes. Signal handlers registered on the parent process (SIGTERM, SIGINT) won't fire during this time in the expected way.

The repo's established pattern for long-running subprocesses is async `spawn` wrapped in a Promise (see `src/lib/flyctl.ts:112`). Using async `spawn` with `stdio: 'inherit'` achieves the same TTY-passthrough behavior without blocking.

#### Functionality — Exit code 130 treated as a hard error (`setup.ts:108-110`)

```typescript
if (exitCode !== 0) {
  throw new CLIError(`PostHog wizard exited with code ${exitCode}.`);
}
```

If the user presses Ctrl+C inside the wizard, it exits with code 130 (SIGINT). The current code surfaces this as *"PostHog wizard exited with code 130"* — a confusing error for an intentional cancellation. Consider special-casing `130` (and `signal === 'SIGINT'` on the `spawnSync` result) to print a friendlier "Setup cancelled." message and exit cleanly.

---

### Information

#### Design — `runConnectFlow` return value is silently discarded (`setup.ts:148`)

```typescript
await runConnectFlow(projectId, token, startResult.authorizeUrl, opts);
return 'newly-connected';
```

`runConnectFlow` returns `Promise<PosthogConnectionResponse>` but the result is unused. This is harmless — the connection object is no longer needed — but the return type is now misleading. Consider changing `runConnectFlow` to return `Promise<void>` to signal intent clearly.

#### Security — `@posthog/wizard@latest` is unpinned

The command always downloads and executes the current `@posthog/wizard@latest` at runtime. This is a deliberate design choice (wizard manages its own upgrades) and acceptable for delegating to a first-party tool, but worth acknowledging: if PostHog's npm account were compromised, any published `@latest` would run in the user's environment with their full filesystem access. Pinning to a minor range (e.g. `@posthog/wizard@^0`) rather than `@latest` would limit blast radius from a hypothetical supply-chain incident.

---

## Verdict

**`request_changes`** — two Critical findings (missing tests for new behavior; JSON mode always fails). Suggestion items can be addressed in a follow-up if preferred, but the Critical items must be resolved before merge.

jwfing (commented):
## Summary

Clean, well-scoped refactor that correctly delegates SDK wiring to `@posthog/wizard` while keeping the InsForge dashboard connection step; the code is leaner and the tests cover the main flows.

## Requirements context

No `/docs/superpowers/` directory exists in this repo — assessing against the PR description and code intent alone.

## Findings

### Critical

_(none)_

### Suggestion

**[Software Engineering — Test quality] JSON mode test has an unintentional first-run failure**
`src/commands/posthog/setup.test.ts:234-263`

The `'skips wizard, emits JSON with wizardCommand'` test calls `await runSetup(['--skip-browser'])` before the JSON-mode run, but never configures `spawnSyncMock`. After `beforeEach`'s `mockReset()`, `spawnSyncMock()` returns `undefined`, so `wizardResult.error` throws a `TypeError`, `handleError` calls `process.exit(1)`, and the first run silently fails. `spawnSyncMock.mockClear()` then hides the evidence before the real assertion. The final assertions still validate the correct behaviour, but the first call is dead weight and may confuse future maintainers.

Fix: either add `spawnSyncMock.mockReturnValue(spawnOk())` before the first call, or simply drop the first `runSetup` call entirely — the API mocks set at lines 236–242 remain configured for the second (JSON-mode) invocation without it.

---

**[Functionality — Error message accuracy] `forbidden` from the sanity-check `fetchPosthogConnection` reports "not-connected"**
`src/commands/posthog/setup.ts:167-173`

When `startPosthogCliFlow` returns `{ type: 'connected' }` but the subsequent `fetchPosthogConnection` sanity-check returns `{ kind: 'forbidden' }`, the code throws:

```
'cli-start reported connected, but /connection returned not-connected. Try again, or check the dashboard.'
```

A permissions error is not the same as not-connected — the user will be confused about why retrying would help. Worth handling `kind === 'forbidden'` distinctly with a message pointing at project access, consistent with how `pollPosthogConnection` handles it (exit code 5, explicit forbidden message at `posthog.ts:181`).

---

**[Functionality — Result accuracy] SIGINT-via-signal reports `wizardExitCode: 1` instead of `130`**
`src/commands/posthog/setup.ts:128-138`

```ts
const exitCode = wizardResult.status ?? 1;
if (wizardResult.signal === 'SIGINT' || exitCode === 130) { ... }
```

When the wizard is killed by SIGINT (Node sets `status: null`, `signal: 'SIGINT'`), `exitCode` resolves to `1` via the `?? 1` fallback. The resulting `SetupResult` then carries `wizardExitCode: 1`, even though the conventional exit code for SIGINT is `130`. Consumers that inspect `wizardExitCode` to distinguish cancellation from a real wizard failure will misread this. Consider `const exitCode = wizardResult.signal === 'SIGINT' ? 130 : (wizardResult.status ?? 1)`.

### Information

**`env: process.env` in `spawnSync` is redundant**
`src/commands/posthog/setup.ts:120-123`

`spawnSync` inherits `process.env` by default when `env` is omitted; passing it explicitly has no effect. Safe to drop.

## Verdict

**approved** _(informational — no Critical findings; human approval via the approve flow is still required)_

Three Suggestion-level items worth addressing before merge, but none are blocking. The JSON mode test issue is the most important to fix — it should either configure `spawnSyncMock` or remove the redundant first `runSetup` call so the test reads cleanly.

jwfing (approved):
LGTM, approved.

PR #137, opened 2026-05-21
feat(cli): add --marketplace flag to create

CarmenDou:
---
## Summary by cubic
Adds `--marketplace <slug>` to `create` to install marketplace templates by slug with early validation and a success-gated download counter. The marketplace path now reuses the GitHub template workflow and only reports a download on successful installs.

- **New Features**
  - `--marketplace <slug>` installs a template without the picker; mutually exclusive with `--template`.
  - Validates slugs (lowercase letters, numbers, hyphens) at action entry before auth/project creation.
  - Reuses `downloadGitHubTemplate` to clone from the canonical InsForge templates repo on `main` (no env-var overrides), copy files, seed `.env.local`, and run `migrations/db_init.sql` if present; returns a boolean for success.
  - Sends a POST to `${apiUrl}/templates/v1/<slug>/downloads` only when the download succeeds (slug URL-encoded); network/non-2xx errors are ignored. PostHog payload remains unchanged.

- **Migration**
  - Example: `insforge create my-app --marketplace chatbot`

<sup>Written for commit 150bb96ab6396856a41a5855689279ec069cafab. Summary will update on new commits. <a href="https://cubic.dev/pr/InsForge/CLI/pull/137?utm_source=github">Review in cubic</a></sup>

> [!NOTE]
> ### Add `--marketplace` flag to the `create` command for installing marketplace templates
> - Adds `--marketplace <slug>` to the `create` CLI command, which validates the slug against `SAFE_MARKETPLACE_SLUG` (`^[a-z0-9][a-z0-9-]{0,99}$`) before authentication or project creation.
> - Mutually excludes `--marketplace` and `--template`; providing both throws a `CLIError`.
> - On successful download, fires a best-effort POST to `/templates/v1/<slug>/downloads` via the new `reportMarketplaceDownload` function, which swallows all errors.
> - Changes `downloadGitHubTemplate` to return `Promise<boolean>` (true on success, false on failure) to gate the download counter ping.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 150bb96.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Scaffold projects from the marketplace using a new --marketplace flag
  * Marketplace scaffolding downloads templates, configures environment files without overwriting existing local env, and runs database initialization when present
  * --marketplace and --template are mutually exclusive; marketplace selection skips interactive picker

* **Tests**
  * Added comprehensive tests covering marketplace download/validation, error handling, and telemetry/reporting behavior (fire-and-forget reporting)

[Review Change Stack](https://app.coderabbit.ai/change-stack/InsForge/CLI/pull/137?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/commands/create.marketplace.test.ts
+++ b/src/commands/create.marketplace.test.ts
@@ -1,0 +1,86 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { readFileSync } from 'node:fs';
+import { join, dirname } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { reportMarketplaceDownload } from './create';
+
+const here = dirname(fileURLToPath(import.meta.url));
+const createSource: string = readFileSync(join(here, 'create.ts'), 'utf8');
+
+describe('reportMarketplaceDownload', () => {
+  const fetchMock = vi.fn();
+
+  beforeEach(() => {
+    vi.stubGlobal('fetch', fetchMock);
+    fetchMock.mockReset();
+  });
+
+  afterEach(() => {
+    vi.unstubAllGlobals();
+  });
+
+  it('POSTs to /templates/v1/<slug>/downloads on the given apiUrl', async () => {
+    fetchMock.mockResolvedValue({ ok: true, status: 200, json: async () => ({ count: 1 }) });
+
+    await reportMarketplaceDownload('chatbot', 'https://api.insforge.dev');
+
+    const [url, init] = fetchMock.mock.calls[0];
+    expect(url).toBe('https://api.insforge.dev/templates/v1/chatbot/downloads');
+    expect(init).toMatchObject({ method: 'POST' });
+  });
+
+  it('URL-encodes the slug', async () => {
+    fetchMock.mockResolvedValue({ ok: true, status: 200, json: async () => ({ count: 1 }) });
+
+    await reportMarketplaceDownload('weird slug', 'https://api.insforge.dev');
+
+    const [url] = fetchMock.mock.calls[0];
+    expect(url).toContain('weird%20slug');
+  });
+
+  it('swallows network errors (does not throw)', async () => {
+    fetchMock.mockRejectedValue(new Error('ECONNREFUSED'));
+
+    await expect(
+      reportMarketplaceDownload('chatbot', 'https://api.insforge.dev'),
+    ).resolves.toBeUndefined();
+  });
+
+  it('swallows non-2xx responses (does not throw)', async () => {
+    fetchMock.mockResolvedValue({ ok: false, status: 503, json: async () => ({}) });
+
+    await expect(
+      reportMarketplaceDownload('chatbot', 'https://api.insforge.dev'),
+    ).resolves.toBeUndefined();
+  });
+});
+
+describe('--marketplace flag wiring', () => {
+  // These tests read the source of create.ts and assert structural invariants
+  // (presence of mutual-exclusion check, branch ordering, absence of PostHog
+  // payload changes). The full command involves auth + network and isn't unit-
+  // testable without heavy mocking; this layer guards against silent refactors
+  // that would drop the wiring.
+
+  it('asserts the flag is mutually exclusive with --template at action entry', () => {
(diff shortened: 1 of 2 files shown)


jwfing (commented):
## Code Review — `feat(cli): add --marketplace flag to create`

**Summary:** Clean, well-scoped feature addition. The implementation is correct, the security model is sound, and the test coverage is meaningful. No blocking issues found.

---

### Requirements context

No `/docs/superpowers/` directory exists in this repo. Review assessed against the PR description (cubic + Macroscope summaries) and the code itself. The JSDoc comment on `reportMarketplaceDownload` references "per spec §6.3," suggesting an internal spec exists, but it was not available for consultation.

---

### Findings

#### Critical

_(none)_

---

#### Suggestion

**[Functionality] Stale comment references a non-existent function — `create.ts:175–179`**

```ts
// The defense-in-depth check inside
// downloadMarketplaceTemplate also fires, but by the time that runs
// the project is already linked — a bad slug here would otherwise
// leave an orphaned project on the user's account.
```

`downloadMarketplaceTemplate` does not exist in this PR (the marketplace path reuses `downloadGitHubTemplate` instead). The "defense-in-depth check" this comment promises **never fires**. The code is still safe — the early `SAFE_MARKETPLACE_SLUG` validation is the only gate, and `downloadGitHubTemplate` independently rejects missing template directories via `fs.stat` — but the comment actively misleads readers into thinking there is a second slug-validation layer when there isn't. Please remove or rewrite it.

---

**[Software Engineering] `INSFORGE_TEMPLATES_REPO` / `INSFORGE_TEMPLATES_BRANCH` env-vars do apply to marketplace installs — `create.ts:640–649`**

The PR description says marketplace clones "from the canonical InsForge templates repo on `main` (no env-var overrides)," but the code calls the unmodified `downloadGitHubTemplate`, which reads those env vars. In practice this is probably the right behaviour (keeps the dev escape-hatch consistent), but the description and the code contradict each other. If marketplace slugs are expected to always resolve against the production repo (so a developer's `INSFORGE_TEMPLATES_REPO` pointing at a local fork doesn't silently install the wrong template), the marketplace branch should either strip those vars or shadow them with hardcoded values before calling `downloadGitHubTemplate`. If the current behaviour is intentional, update the PR description and the comment at `create.ts:258` ("no env-var overrides").

---

#### Information

**[Software Engineering] Source-parsing structural tests are brittle — `create.marketplace.test.ts:74–107`**

The tests that read `create.ts` as a raw string and compare character indices are a reasonable workaround for a command that is hard to unit-test end-to-end, and they're well-commented. One fragile spot: the regex `/if \(downloaded\)[\s\S]{0,80}reportMarketplaceDownload/` (line ~103) will silently break if more than ~80 characters of intermediate code are inserted between the condition and the call (e.g., a log line). Consider loosening the `{0,80}` upper bound or switching to a line-count heuristic.

---

**[Software Engineering] Ignored return value in `githubTemplates` branch — `create.ts:390–391`**

`downloadGitHubTemplate` now returns `Promise<boolean>`, but the existing call site for built-in GitHub templates ignores the value:

```ts
} else if (githubTemplates.includes(template!)) {
  await downloadGitHubTemplate(template!, projectConfig, json);
```

TypeScript is happy with this, and the behaviour is unchanged, but an explicit `void` cast (`void await downloadGitHubTemplate(...)`) or a comment would signal that the boolean is deliberately discarded here (unlike the marketplace path, which gates on it). Minor, but makes intent clearer.

---

**[Functionality] `captureEvent('template_selected', { template })` will record marketplace slugs as template names**

When `--marketplace chatbot-clone` is passed, the PostHog payload becomes `{ template: 'chatbot-clone', approach: 'template' }`. Any analytics query that treats `template` as a fixed enum will see unexpected values. This is probably acceptable, but worth noting if dashboards filter or group by `template`.

---

### Verdict

**Approved** _(informational — no Critical findings; explicit GitHub approval is a separate human action)_

The implementation is correct and safe. The stale `downloadMarketplaceTemplate` comment is the most important thing to fix before this lands, but it's a documentation issue, not a correctness bug.

jwfing (approved):
LGTM, approved.

PR #156, opened 2026-06-01
INS-305 Expand declarative config apply coverage

Fermionic-Lyu:
## Summary
- Extend insforge.toml support for auth.disable_signup, storage.max_file_size_mb, realtime.retention_days, and schedules.retention_days.
- Keep the original metadata-first config command flow; apply/plan fetch optional endpoint-backed slices only when the TOML includes those sections, while export probes them to emit a fuller file.
- Attach HTTP status/backend error code to CLIError and treat optional endpoint route-level 404s as unsupported by status/code, not rendered message text.
- Document the expanded declarative config surface in README.

## Not included
- Auth email templates are intentionally not managed through insforge.toml.
- OAuth apps, storage bucket lifecycle, realtime channels, deployment environment variables, functions, and secrets remain outside config apply because they involve external resources or lifecycle ambiguity.

## Related
- Skills docs sync: https://github.com/InsForge/insforge-skills/pull/89

## Test plan
- npm test -- src/commands/config/apply.test.ts src/commands/config/plan.test.ts src/commands/config/export.test.ts src/lib/config-metadata.test.ts src/lib/config-capabilities.test.ts
- npm run lint (passes; existing branch command test no-explicit-any warnings remain)
- npm run build
- Local OSS localhost:7130: config plan/apply/plan restore cycle, then config export with storage/realtime/schedules present and deployments.subdomain skipped on self-host

> [!NOTE]
> ## Summary
> - Extend insforge.toml support for auth.disable_signup, storage.max_file_size_mb, realtime.retention_days, and schedules.retention_days.
> - Keep the original metadata-first config command flow; apply/plan fetch optional endpoint-backed slices only when the TOML includes those sections, while export probes them to emit a fuller file.
> - Attach HTTP status/backend error code to CLIError and treat optional endpoint route-level 404s as unsupported by status/code, not rendered message text.
> - Document the expanded declarative config surface in README.
>
> ## Not included
> - Auth email templates are intentionally not managed through insforge.toml.
> - OAuth apps, storage bucket lifecycle, realtime channels, deployment environment variables, functions, and secrets remain outside config apply because they involve external resources or lifecycle ambiguity.
>
> ## Related
> - Skills docs sync: https://github.com/InsForge/insforge-skills/pull/89
>
> ## Test plan
> - npm test -- src/commands/config/apply.test.ts src/commands/config/plan.test.ts src/commands/config/export.test.ts src/lib/config-metadata.test.ts src/lib/config-capabilities.test.ts
> - npm run lint (passes; existing branch command test no-explicit-any warnings remain)
> - npm run build
> - Local OSS localhost:7130: config plan/apply/plan restore cycle, then config export with storage/realtime/schedules present and deployments.subdomain skipped on self-host
> #### Changes since #156 opened
>
> - Bumped version from 0.1.86 to 0.1.87 [213d29c]
> 
>


jwfing (changes requested):
## Code review — Expand declarative config apply coverage

**Summary:** Cleanly factored extension of `config apply/plan/export` to four new keys (`auth.disable_signup`, `storage.max_file_size_mb`, `realtime.retention_days`, `schedules.retention_days`) via a shared `loadConfigState` loader, well-tested at the unit level — but the "skip unsupported backends gracefully" path is fragile and effectively untested, which can regress the existing config commands on older backends.

**Requirements context:** No matching spec/plan found under `docs/specs/` (the only specs there cover the `diagnose` and `db-migrations` commands) — assessed against the PR description and the surrounding code conventions alone. The PR description states two load-bearing requirements I held the code to: (a) "missing or unavailable endpoints are treated as absent and skipped gracefully" and (b) "keep unsupported backend versions on the existing skip path."

---

### Critical

**1. [functionality / software-engineering] The unsupported-endpoint detection is fragile and untested — it can break `apply`/`plan`/`export` on backends that lack the new endpoints.** — `src/lib/config-metadata.ts:70-110`

`loadConfigState` unconditionally probes `/api/storage/config`, `/api/realtime/config`, and `/api/schedules/config` on **every** invocation (even an auth-only apply), and relies on `isOptionalEndpointUnsupported` to swallow the "endpoint doesn't exist" error. That guard is unreliable:

- `err.code === 'NOT_FOUND'` is a **dead branch**. `ossFetch` only ever throws `new CLIError(message)` with no `code` argument (`src/lib/api/oss.ts:161`, `src/lib/errors.ts:3-12`), so `err.code` is always `undefined` here. Detection therefore rests entirely on the message-substring checks.
- The substring checks miss the most likely route-miss shape. `ossFetch` builds `message = err.message ?? err.error ?? "OSS request failed: ${status}"` (`oss.ts:135`). This file's own `isRouteLevel404 = !err.error || err.error === 'NOT_FOUND'` (`oss.ts:144`) shows the backend returns `{ "error": "NOT_FOUND" }` for route-level 404s. When such a body has no `message` field, `message` becomes the literal `"NOT_FOUND"` → lowercased `"not_found"`, which matches **none** of `'oss request failed: 404'`, `'not found'`, `'not available'`, `'not enabled'` (note the underscore vs. space). The error then propagates and the whole command fails.
- Because the probe is unconditional, this isn't limited to the new sections: a user who upgrades the CLI but whose backend predates these endpoints would find `config export/plan/apply` **completely broken**, even when their `insforge.toml` only touches auth — a regression of existing behavior and a direct violation of requirements (a) and (b) above.
- **No test exercises the catch path.** Every test mock (`apply.test.ts`, `export.test.ts`, `plan.test.ts`) returns HTTP 200 — the "unsupported" cases simulate absence via an empty `{}` 200 body, never a 404 throw. So `fetchOptionalJson`/`isOptionalEndpointUnsupported` — the exact fragile logic — has zero coverage.

Recommended fix: detect by HTTP status, not message text. Either have `ossFetch` attach the status code to the `CLIError` (it already reads `res.status` / `err.statusCode`) and check `err.statusCode === 404` in `fetchOptionalJson`, or do a raw `fetch` in `loadConfigState` and branch on `res.status === 404`. Then add a regression test where the optional endpoint throws a 404 `CLIError` in both the `{ "error": "NOT_FOUND" }` and empty-body shapes (see Suggestion 4).

---

### Suggestion

**2. [functionality / performance] Probe only the endpoints the operation needs.** — `src/lib/config-metadata.ts:70-88`
`loadConfigState` fetches all three optional slices on every `apply`/`plan`, regardless of which sections the TOML contains. For an auth-only apply that's three wasted round-trips (they're parallelized via `Promise.all`, which is good, so it's latency not serialization), and it widens the blast radius of the Critical above. Consider probing only for sections present in `file` for `apply`/`plan`. `export` legitimately needs all three.

**3. [functionality] Confirm the HTTP verbs match the backend routes.** — `src/commands/config/apply.ts:247-266`
`storage` applies with `PUT`, while `realtime`/`schedules` use `PATCH`. That may be correct, but the asymmetry is worth a confirming note: the capability probe is a `GET`, so a write-verb mismatch (e.g. PUT to a PATCH-only route) wouldn't be caught by the gate and would only fail at apply time.

**4. [software-engineering] Add a regression test for the graceful-skip path.** A test where an optional endpoint throws a 404 `CLIError` (both the structured `{ "error": "NOT_FOUND" }` and the no-body `"OSS request failed: 404"` shapes) would lock in the behavior the PR promises and would have caught the Critical. Current tests only cover the 200-with-empty-body case.

---

### Information

**5. [functionality] `max_file_size_mb` upper bound is hard-coded to 200.** — `src/lib/config-schema.ts` (`validateStorage`)
If the backend's actual cap differs or changes, the CLI will reject otherwise-valid configs client-side. Consider documenting where `200` comes from, or letting the backend be the authority on the limit.

**6. [functionality] `retention_days = 0` is overloaded to mean "disable cleanup" (wire `null`).** — `src/lib/config-diff.ts` (`normalizeRetentionDays`), README
Reasonable given TOML has no `null` literal, and it's documented in the README and code comments — just noting that `0` reads naturally as "delete immediately / keep 0 days," so the overload is a mild footgun. No change requested.

**7. [software-engineering] Minor indentation drift** in the reformatted expected object in `src/lib/config-metadata.test.ts` (around the `password`/`disable_signup`/`smtp` block) — valid but inconsistently indented. Cosmetic.

---

### Security
No security-relevant concerns. New writes go through the existing `ossFetch` (Bearer api_key); all new TOML inputs are validated with type/range checks (`config-schema.ts`); analytics records only section path names, not values, so no secrets/PII are newly logged; no new dependencies. SMTP/secret handling is unchanged.

### Performance
No N+1 or event-loop blocking. The only note is the three unconditional optional-endpoint probes per invocation (parallelized) — see Suggestion 2.

---

One Critical item (the fragile, untested unsupported-endpoint detection that can regress existing config commands on older backends). The rest of the change is well-structured and the unit-test coverage for the happy paths is solid. Resolving the Critical — status-based 404 detection plus a regression test — should make this mergeable.

jwfing (changes requested):
## Review: Expand declarative config apply coverage

**Summary:** Solid, well-factored extension of `config apply/plan/export` to four new declarative keys, but the new "optional endpoint" 404 detection matches a message string that real InsForge backends never emit, which will make all three commands throw against the exact older backends the graceful-skip path is meant to support.

### Requirements context
No `/docs/superpowers/` directory exists in this repo, and nothing under `docs/specs/` (the three existing specs cover `diagnose` and `db migrations`) matches declarative config apply — so I assessed against the PR description/body and the repo's `DEVELOPMENT.md` + `.claude/skills/cli-development` conventions, plus the existing config-* code as the source of intent.

---

### 🔴 Critical

**[Functionality] Optional-endpoint 404 detection matches a string the backend never sends → `plan`/`apply`/`export` crash on older backends — `src/lib/config-metadata.ts:98-102`, `src/lib/api/oss.ts:127-161`**

`fetchOptionalJson` swallows a 404 only when `isOptionalEndpointUnsupported(err)` is true, which requires the `CLIError.message` to equal `oss request failed: 404` (or start with `oss request failed: 404\n`):

```ts
return message === 'oss request failed: 404' || message.startsWith('oss request failed: 404\n');
```

But that string is the CLI's *client-side fallback*, produced by `ossFetch` only when the 404 body has **no** `message` and **no** `error` field (`oss.ts:135` — `err.message ?? err.error ?? 'OSS request failed: ${res.status}'`). A real InsForge route-level 404 returns a JSON body `{ error: 'NOT_FOUND' }` — this is pinned by the repo's own existing test `src/lib/api/oss.test.ts:69` and by the route-level handling in `oss.ts:144` (`const isRouteLevel404 = !err.error || err.error === 'NOT_FOUND'`). For that body, `ossFetch` builds `CLIError('NOT_FOUND')`, so `isOptionalEndpointUnsupported` sees `'not_found'`, returns **false**, and the error propagates out of `loadConfigState` — failing the whole command.

Impact: before this PR, `plan/apply/export` hit only `/api/metadata` (always present). This PR adds three probes (`/api/storage/config`, `/api/realtime/config`, `/api/schedules/config`) to **every** invocation (`config-metadata.ts:74-78`). Any backend lacking even one of them now throws instead of skipping — a regression that directly contradicts the PR's stated behavior ("missing or unavailable endpoints are treated as absent and skipped gracefully" / "keep unsupported backend versions on the existing skip path"). The graceful path only fires for a bare-bodied/non-JSON 404 (e.g. an nginx/proxy 404), not the application's own `NOT_FOUND` response.

Recommended fix: stop string-matching the rendered message. `ossFetch` already parses `statusCode` (`oss.ts:131-132`) and knows `res.status` — attach the HTTP status to `CLIError` and gate on `err.status === 404` in `isOptionalEndpointUnsupported`. (Cheap and also removes the brittleness flagged below.) Please also verify the actual 404 body shape these three endpoints return on a pre-feature backend before merging.

---

### 🟡 Suggestion

**[Software engineering / testing] The graceful-skip test feeds an error the backend can't produce — `src/lib/config-state-loader.test.ts:24-36`**
The "treats route-level missing optional endpoints as unsupported" test throws `new CLIError('OSS request failed: 404')`, i.e. the synthetic fallback, so it passes while the real `{error:'NOT_FOUND'}` path (the Critical above) goes untested — the test gives false green on the central new behavior. Add a case that drives a realistic route-level 404, ideally end-to-end through `ossFetch`'s error mapping (mock `fetch` returning `Response(JSON.stringify({error:'NOT_FOUND'}), {status:404})`, as `oss.test.ts:69` does) rather than hand-constructing the `CLIError`. That test would have caught this.

**[Project convention] Agent-skills sync not mentioned — `DEVELOPMENT.md` §3**
This PR adds a new user-facing declarative surface (`auth.disable_signup`, `storage.max_file_size_mb`, `realtime.retention_days`, `schedules.retention_days`). `DEVELOPMENT.md` §3 asks that command/flag/surface changes be mirrored in the `insforge-cli` skill in `InsForge/agent-skills` and that the skill PR be cross-referenced here. README is updated (good), but there's no mention of an agent-skills update — please confirm whether the skill needs the new keys, or note that it doesn't.

---

### 🔵 Information

**[Design] Root cause of the brittle matching: `CLIError` carries no HTTP status — `src/lib/api/oss.ts:127-161`**
`statusCode` is read off the body at `oss.ts:131-132` and `res.status` is in scope, but neither is attached to the thrown `CLIError`, forcing consumers to recover intent by string-matching `.message`. This same pattern already appears in `config-secrets.ts` (matching for `SECRET_NOT_FOUND`). Carrying `status` on `CLIError` would let all of them gate on a code instead of a localized message — worth doing alongside the Critical fix.

**[Functionality] `asNumber` silently drops non-integer values — `src/lib/config-metadata.ts:223-229`**
`asNumber`/`asRetentionDays` require `Number.isInteger`, so a backend that returns `maxFileSizeMb`/`retentionDays` as a float or numeric string degrades to "unsupported/skipped" rather than surfacing a parse warning. Low blast radius given the integer-only schema, but worth a comment so it isn't read as a capability gap later.

**[Consistency] Mixed HTTP verbs across the new endpoints — `src/commands/config/apply.ts:247-266`**
`storage` uses `PUT /api/storage/config` while `realtime`/`schedules` use `PATCH /api/{realtime,schedules}/config`. Presumably matches each backend route's contract — just flagging to confirm it's intentional, not a copy-paste drift.

---

### Dimension coverage notes
- **Security:** No findings. SMTP password stays an `env()` ref, is force-resent and resolved at PUT time, and is never logged or round-tripped into TOML (`config-diff.ts:449-465`, `apply.ts:229-237`); the new `disable_signup` toggle is capability-gated like the rest; analytics properties stay limited to counts/section names per `DEVELOPMENT.md` §2. New endpoints reuse the existing Bearer-key `ossFetch` path — no new injection surface.
- **Performance:** No findings. The three new probes run via `Promise.all` after `/api/metadata` (`config-metadata.ts:74-78`); +3 parallel requests per invocation, documented in the PR body, no N+1 or hot-path concern.

---

One Critical (the 404-detection regression) blocks merge. The structure, type-safety (exhaustive `DiffChange` checks), round-trip discipline, and the rest of the test coverage are good — once the optional-endpoint detection is keyed to the HTTP status (and a realistic 404 test is added), this should be in good shape.

Fermionic-Lyu:
Addressed the review items in `4a028b8`:

- Removed the `loadConfigState` / config-state wrapper path and restored metadata-first command flow.
- `apply` and `plan` now only probe optional config endpoints when the TOML includes the corresponding section; auth-only runs do not touch storage/realtime/schedules endpoints.
- `ossFetch` now attaches backend error code and HTTP status to `CLIError`; optional route-miss handling keys off `statusCode === 404` plus structured code (`undefined` or `NOT_FOUND`), not message text.
- Added regression coverage for both route-miss shapes: `{ error: "NOT_FOUND" }` and empty-body 404, plus a resource-level 404 case that still surfaces.
- Synced agent skill docs in https://github.com/InsForge/insforge-skills/pull/89 and linked it in this PR description.

I saw the new cubic P3 about centralizing the duplicated optional-endpoint helpers; I am intentionally leaving those local for now to keep this PR minimal and avoid reintroducing a wrapper abstraction after reverting the state-loader approach.

jwfing (commented):
## Review: Expand declarative config apply coverage

**Summary:** Cleanly extends `insforge.toml` to cover `auth.disable_signup`, `storage.max_file_size_mb`, `realtime.retention_days`, and `schedules.retention_days`, with consistent capability-gating and solid test coverage — no blocking issues.

### Requirements context
No `/docs/superpowers/` directory exists in this repo, and the specs under `docs/specs/` (diagnose, db-migrations) do not match this PR's scope — **no matching spec/plan found; assessed against the PR description, the existing config-apply architecture, and `DEVELOPMENT.md` conventions.** The change is well-scoped: every touched line serves the four stated new sections plus the supporting `CLIError` status/code plumbing.

---

### Critical
(none)

No correctness bugs, security holes, data-loss risks, or broken existing behavior found in the new code.

---

### Suggestions

**1. Software engineering — duplicated optional-endpoint helpers (DRY).**
`fetchOptionalConfig<T>()` and `isMissingOptionalEndpoint()` are copied verbatim into three files: `src/commands/config/apply.ts:190-206`, `src/commands/config/plan.ts:107-123`, and `src/commands/config/export.ts:114-130`. Since `EndpointConfigResponses` and the metadata projection already live in `src/lib/config-metadata.ts`, hoisting these two helpers into a shared module would keep the route-level-404 detection logic in one place (it must stay in sync with the `isRouteLevel404` rule in `src/lib/api/oss.ts:144`). Non-blocking, but it's three copies of a security/compat-sensitive predicate.

**2. Performance — optional config fetches are sequential in apply/plan.**
`export.ts:63-67` fetches the three optional config slices with `Promise.all`, but `apply.ts:51-65` and `plan.ts:43-57` `await` them one at a time. When a TOML declares all of `storage` + `realtime` + `schedules`, that's three sequential round-trips that could be parallelized like `export` does. Low blast radius (cold CLI path, ≤3 calls, gated on section presence), but worth aligning for consistency.

**3. Software engineering — new schema validators are only covered indirectly.**
`validateStorage` and `validateRetentionSection` (`src/lib/config-schema.ts:116-163`) have no dedicated `config-schema.test.ts`. They are exercised through `parseConfigToml` in `config-toml.test.ts` (the `201` over-max and `-1` negative rejection cases are good), but the valid boundary values (`max_file_size_mb = 1` and `= 200`, `retention_days = null` literal) aren't directly asserted. Coverage is reasonable as-is; adding the boundaries would harden the range checks against future edits.

---

### Information

**1. Functionality — `EMPTY_STORAGE_CONFIG.max_file_size_mb = 50` default** (`src/lib/config-diff.ts:524-526`). This fallback only surfaces as the `from` value in plan output for backends that don't expose `/api/storage/config` — and in exactly that case the change is capability-skipped, so it never drives an actual PUT. Harmless, and it mirrors the existing `EMPTY_PASSWORD_POLICY` pattern; just be aware the displayed `from: 50` is a CLI-side guess, not a real backend read, when the section is unsupported.

**2. Software engineering — `CLIError` widening is safely additive** (`src/lib/errors.ts:8`, `src/lib/api/oss.ts:161`). `ossFetch` now passes `err.error` as `code` and `res.status` as `statusCode` on every thrown `CLIError` (previously both were undefined). I checked the codebase: all other `.code ===` / `.statusCode` reads are on Node `ErrnoException`s (`ENOENT`/`EEXIST`/numeric exit codes), never on `ossFetch`'s `CLIError`, so no existing branch changes behavior. The `--json` error output now also carries the backend's error code, which is an improvement.

**3. Security — no concerns.** The new sections carry only integers/booleans; no new user input reaches SQL/shell. SMTP secret handling (env() refs, force-resend, never round-tripping the value) is pre-existing and untouched. Analytics `sections_changed` records only `section.key` paths (`changePath`), never values — no PII/secret leakage.

**4. Round-trip correctness verified.** The `retention_days = 0` ⇄ wire-`null` mapping (`normalizeRetentionDays`, `config-diff.ts:420-424`; `renderRetentionFields`, `config-toml.ts:161-166`) is symmetric and tested, and `config-format.ts`'s generic `formatChange` fallback (`:61`) renders the new variants without modification. The `0 disables retention` semantics is documented in the README.

---

### Verdict
**`approved`** (informational — a human still gives the explicit GitHub approval). Zero Critical findings. The implementation matches the PR's stated scope, capability-gates each new section independently, handles optional-endpoint 404s correctly, and is backed by thorough apply/plan/export/diff/metadata tests. The suggestions above are non-blocking polish.

jwfing (approved):
LGTM - approved.

PR #163, opened 2026-06-10
INS-347 feat(deployments): respect .vercelignore in deployments deploy

Fermionic-Lyu:
## Summary

`deployments deploy` now honors a `.vercelignore` file in the deploy source directory, letting developers maintain a deploy ignore list. Since deployments run on Vercel underneath, we reuse the existing `.vercelignore` convention instead of introducing a new file — projects migrating from direct Vercel deploys keep working unchanged.

- New `src/commands/deployments/ignore-file.ts`: loads `.vercelignore` via the `ignore` package (gitignore-spec matcher, zero deps). Missing file → no-op; unreadable file → clear `CLIError`.
- Applied in **both** upload paths: the direct-upload scan (`collectDeploymentFiles`, which also skips recursing into ignored directories) and the legacy zip fallback (`createZipBuffer`). `create`'s first-deploy flow reuses `deployProject()` and benefits automatically.
- Full gitignore syntax including `!` negation. Built-in safety excludes (`.env`, `.git`, `node_modules`, …) always win and cannot be re-included. `.vercelignore` itself is never uploaded.
- Non-JSON mode prints `Applying .vercelignore (N patterns)` before the scan.
- Telemetry: `has_ignore_file: boolean` added to the deploy event (file presence only, never pattern contents).
- README: documented in the `deployments deploy` section.

## Scope

Only the `deployments` tree — `compute deploy` and `functions deploy` have separate flows and are unaffected. Version is not bumped here; bump on release per the usual workflow.

## Cross-reference

Companion docs update to the `insforge-cli` skill: InsForge/insforge-skills#94 (merge that one after or together with this).

## Test plan

- [x] 8 new vitest cases in `src/commands/deployments/ignore-file.test.ts` (parsing, comments/CRLF, negation, directory excludes, built-ins not overridable, no-file passthrough) against real temp dirs
- [x] Full suite: 429 passed
- [x] `npm run lint` clean (0 errors)
- [x] `npm run build` + `deployments deploy --help` smoke test

## Summary by CodeRabbit

* **New Features**
  * Add support for a deploy ignore file (.vercelignore) to exclude files/directories using .gitignore syntax (including negation); the ignore file itself is never uploaded.

* **Documentation**
  * Updated Deployments docs with instructions and a sample .vercelignore.

* **Bug Fixes / Behavior**
  * Built-in excludes (e.g., node_modules, .git, .env) remain enforced and cannot be re-included.

* **Tests**
  * Added tests covering ignore-file parsing and deployment exclusion behavior.

* **Chores**
  * Added runtime dependency for ignore pattern handling.

> [!NOTE]
> ### Add `.vercelignore` support to exclude files from deployment uploads
> - Introduces [ignore-file.ts](https://github.com/InsForge/CLI/pull/163/files#diff-0cec94c0d3347d1fd9cf228f3374d4d08cb02e3bca5b243613f02ddad5a0b339) with `loadDeployIgnore`, which reads `.vercelignore` from the source directory and returns an `ignore`-library matcher (or `null` if absent).
> - Applies the matcher in both direct (`collectDeploymentFiles`) and legacy (`createZipBuffer`) upload paths to skip matched files and directories.
> - The `.vercelignore` file itself is always excluded via `EXCLUDE_PATTERNS`; built-in excludes cannot be overridden by negation patterns.
> - Logs the ignore filename and pattern count when a `.vercelignore` is detected, and records a `has_ignore_file` telemetry flag on deploy.
>
> 
> #### Changes since #163 opened
>
> - Exported `createZipBuffer` function from the deployments deploy module [4f2cd77]
> - Implemented .vercelignore pattern filtering in deployment zip creation [4f2cd77]
> - Added error handling for unreadable .vercelignore file cases [4f2cd77]
> - Created test utilities and test suites for .vercelignore functionality [4f2cd77]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized c82514a.</sup>
> 
>
>

### DIFF
--- a/README.md
+++ b/README.md
@@ -496,6 +496,15 @@ npx @insforge/cli deployments deploy ./my-app
 npx @insforge/cli deployments deploy --env '{"API_URL": "https://api.example.com"}'
 ```
 
+To exclude files from the upload, add a `.vercelignore` file to the deploy directory. It uses `.gitignore` syntax (including `!` negation) and is applied on top of the built-in excludes (`node_modules`, `.git`, `.env`, etc. — these always stay excluded and cannot be re-included).
+
+```gitignore
+# .vercelignore
+*.md
+drafts/
+!IMPORTANT.md
+```
+
 #### `npx @insforge/cli deployments list`
 
 List all deployments.
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,6 +14,7 @@
         "archiver": "^7.0.1",
         "cli-table3": "^0.6.5",
         "commander": "^13.1.0",
+        "ignore": "^7.0.5",
         "open": "^10.1.0",
         "picocolors": "^1.1.1",
         "posthog-node": "^5.28.9",
@@ -1585,16 +1586,6 @@
         "typescript": ">=4.8.4 <6.0.0"
       }
     },
-    "node_modules/@typescript-eslint/eslint-plugin/node_modules/ignore": {
-      "version": "7.0.5",
-      "resolved": "https://registry.npmjs.org/ignore/-/ignore-7.0.5.tgz",
-      "integrity": "sha512-Hs59xBNfUIunMFgWAbGX5cq6893IbWg4KnrjbYwX3tx0ztorVgTDA6B2sxf8ejHJ4wz8BqGUMYlnzNBer5NvGg==",
-      "dev": true,
-      "license": "MIT",
-      "engines": {
-        "node": ">= 4"
-      }
-    },
     "node_modules/@typescript-eslint/parser": {
       "version": "8.56.1",
       "resolved": "https://registry.npmjs.org/@typescript-eslint/parser/-/parser-8.56.1.tgz",
@@ -2642,6 +2633,16 @@
         "url": "https://opencollective.com/eslint"
       }
     },
+    "node_modules/eslint/node_modules/ignore": {
+      "version": "5.3.2",
+      "resolved": "https://registry.npmjs.org/ignore/-/ignore-5.3.2.tgz",
+      "integrity": "sha512-hsBTNUqQTDwkWtcdYI2i06Y/nUBEsNEDJKjWdigLvegy8kDuJAS8uRlpkkcQpyEXL0Z/pjDy5HBmMjRCJ2gq+g==",
+      "dev": true,
+      "license": "MIT",
+      "engines": {
+        "node": ">= 4"
+      }
+    },
     "node_modules/espree": {
       "version": "11.1.1",
       "resolved": "https://registry.npmjs.org/espree/-/espree-11.1.1.tgz",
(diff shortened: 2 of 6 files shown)


jwfing (commented):
## Summary

Solid, well-scoped implementation of `.vercelignore` support for `deployments deploy` — correct use of the `ignore` package (verified against current node-ignore v7 docs, including the trailing-slash directory convention), applied in both upload paths, with good test coverage on the direct-upload flow. No blocking issues.

## Requirements context

No matching spec/plan found — `docs/specs/` contains designs for the diagnose command and db migrations only (there is no `/docs/superpowers/` directory in this repo). Assessed against the PR description, which the implementation fully matches: both upload paths covered, built-ins non-overridable, `.vercelignore` never uploaded, presence-only telemetry, README updated.

## Findings

### Critical

(none)

### Suggestion

1. **[Software engineering — test gap] The legacy zip path is untested, and its directory semantics differ subtly from the walk path.** `createZipBuffer` (`src/commands/deployments/deploy.ts:156-177`) is exercised by zero tests (all 8 new tests go through `collectDeploymentFiles`). The subtlety: archiver entry names for directories don't carry a trailing slash, so a directory-only pattern like `drafts/` won't match the directory entry itself — files beneath it are still correctly excluded (`drafts/wip.txt` matches), but an empty directory entry may survive in the zip, and any future divergence between the two paths won't be caught. Consider exporting `createZipBuffer` for one integration test mirroring the `collectDeploymentFiles` cases (unzip the buffer or list archive entries).

2. **[Software engineering — test gap] The unreadable-file → `CLIError` branch is untested.** `src/commands/deployments/ignore-file.ts:24-31` wraps non-ENOENT read failures in a `CLIError`, but no test covers it. A test using a directory named `.vercelignore` (EISDIR, portable, no chmod flakiness) would lock in the error message.

### Information

1. **[Functionality — minor duplication] `.vercelignore` presence is detected twice.** The command action stats the file separately for telemetry (`src/commands/deployments/deploy.ts:474-478`) while `deployProject` loads it again internally (`deploy.ts:426`). Harmless (worst case is a TOCTOU mismatch in a boolean telemetry field), but surfacing `hasIgnoreFile` from `DeployProjectResult` would keep one source of truth — especially since `create.ts:484` also calls `deployProject` and gets no telemetry flag at all.

2. **[Functionality] Negation into an excluded directory matches git semantics on the direct path.** `collectDeploymentFiles` skips recursion into ignored directories (`deploy.ts:129-131`), so `docs/` + `!docs/keep.md` keeps `keep.md` excluded — same as git ("it is not possible to re-include a file if a parent directory is excluded"). Noting it as intentional-looking behavior worth keeping in mind if users report "my negation doesn't work".

### Dimension check

- **Software engineering**: covered above. Otherwise good — ESM imports with `.js` extensions, `CLIError` for user-facing failures, vitest against real temp dirs with proper cleanup, all consistent with repo conventions. The aggressive-negation test (`!**`, nested `node_modules`, nested `.env`) is a nice safety net for the built-in-excludes guarantee.
- **Functionality**: implementation matches the PR description; check ordering (built-ins first at `deploy.ts:125`, then `.vercelignore` at `deploy.ts:129`) makes built-ins genuinely non-overridable; Windows paths handled via `normalizeRelativePath`; the empty-string guard before `ignores()` in the zip callback avoids node-ignore's invalid-path throw; info log is correctly gated off in JSON mode (spinner is null).
- **Security**: no security-relevant concerns. Only a local file read from the user's own deploy dir; telemetry records presence only (boolean), never pattern contents; no new input reaches shell/SQL/HTTP. New dependency `ignore@^7.0.5` is reputable (MIT, used by ESLint), caret-pinned consistent with the repo's other deps, and the lockfile changes are coherent (promoted from dev-transitive to direct prod dep).
- **Performance**: no concerns. One file read per deploy, O(patterns) matcher check per entry, and skipping recursion into ignored directories is a net win on large trees.

## Verdict

**approved** (informational — explicit GitHub approval remains a human action via the approve flow). Zero Critical findings; the two Suggestions are non-blocking test-coverage improvements.

jwfing (approved):
LGTM - approved.

PR #168, opened 2026-06-16
fix(storage): set content type on CLI uploads instead of octet-stream

Fermionic-Lyu:
## Why

Linear [INS-358](https://linear.app/insforge/issue/INS-358): storage uploads default to `application/octet-stream` instead of the actual file type.

Verified against source that the **CLI is the affected path, on every upload**. `storage upload` built the body with `new Blob([fileContent])` and never set a type ([`src/commands/storage/upload.ts:33`](https://github.com/InsForge/CLI/blob/main/src/commands/storage/upload.ts)). A typeless `Blob` serializes its multipart part as `application/octet-stream`, and the backend trusts the client's value — it stores `file.mimetype || null` and the S3 provider falls back to `application/octet-stream`; it never infers from the extension. So every CLI-uploaded object got the wrong MIME type.

## What

- Infer the MIME type from the file's extension and set it on the blob, so it reaches the server via the multipart part. This is the right place to infer — the CLI reads a real file off disk (same approach `aws s3 cp` uses).
- Add a `--content-type <type>` flag to override the inferred value.
- New `src/lib/mime.ts` helper (small self-contained lookup, no new runtime dependency) + unit tests.
- Unknown extensions still fall back to `application/octet-stream`, so there is **no regression**.

```
storage upload ./cover.png --bucket images                  # → image/png
storage upload ./data.bin  --bucket files --content-type application/json
```

## Test

- `vitest run src/lib/mime.test.ts` ✅ (case-insensitivity, paths, unknown/missing extensions)
- `npm run lint` ✅ (0 errors) · `npm run build` ✅

---
## Summary by cubic
CLI storage uploads now send the correct MIME type instead of always `application/octet-stream`, fixing INS-358. Also bumps `@insforge/cli` to 0.1.91.

- **Bug Fixes**
  - Infer type from the basename’s extension and set it on the multipart Blob; fall back to `application/octet-stream` for unknown/missing extensions.
  - Enforce precedence: override > inferred > fallback; ignore empty/whitespace `--content-type`.

- **New Features**
  - Adds `--content-type <type>` to override the inferred type.
  - Introduces `src/lib/mime.ts` and `resolveUploadContentType` with unit tests; no new runtime dependencies.

<sup>Written for commit 5ca75b3d09ed1560796fb9e5f28580511c4bc278. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/CLI/pull/168?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Set MIME content type on `storage upload` CLI command based on file extension
> - Adds a `--content-type` flag to the `storage upload` command; if omitted, the type is inferred from the file extension via a new `mimeTypeFromName` utility in [mime.ts](https://github.com/InsForge/CLI/pull/168/files#diff-53bf5ead157b92bb4808bad227054fbe0d6d503f08054423cb77683e583c97d5), falling back to `application/octet-stream`.
> - `mimeTypeFromName` covers common extensions across images, documents, text/data, archives, audio, video, and fonts.
> - The resolved type is passed to the `Blob` constructor so the multipart `file` part carries the correct `Content-Type` instead of always sending `application/octet-stream`.
>
> 
> #### Changes since #168 opened
>
> - Modified `mimeTypeFromName` utility to extract file extensions from the basename only [77bb5bc]
> - Introduced `resolveUploadContentType` utility and integrated it into storage upload command [77bb5bc]
> - Added test coverage for content type resolution and MIME type inference [77bb5bc]
> - Bumped package version from 0.1.90 to 0.1.91 [5ca75b3]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0742071.</sup>
> 
>
> 

## Summary by CodeRabbit

* **New Features**
  * Added `--content-type` flag for storage uploads, allowing explicit MIME type specification or automatic detection from file extension, with `application/octet-stream` as fallback.

* **Chores**
  * Version updated to 0.1.91.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@insforge/cli",
-  "version": "0.1.90",
+  "version": "0.1.91",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@insforge/cli",
-      "version": "0.1.90",
+      "version": "0.1.91",
       "license": "Apache-2.0",
       "dependencies": {
         "@clack/prompts": "^0.9.1",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@insforge/cli",
-  "version": "0.1.90",
+  "version": "0.1.91",
   "description": "InsForge CLI - Command line tool for InsForge platform",
   "type": "module",
   "bin": {
--- a/src/commands/storage/upload.test.ts
+++ b/src/commands/storage/upload.test.ts
@@ -1,0 +1,23 @@
+import { describe, it, expect } from 'vitest';
+import { resolveUploadContentType } from './upload.js';
+
+describe('resolveUploadContentType', () => {
+  it('uses an explicit --content-type over inference', () => {
+    expect(resolveUploadContentType('photo.png', 'image/webp')).toBe('image/webp');
+  });
+
+  it('infers from the file extension when no flag is given', () => {
+    expect(resolveUploadContentType('photo.png')).toBe('image/png');
+    expect(resolveUploadContentType('/tmp/report.pdf')).toBe('application/pdf');
+  });
+
+  it('treats an empty or whitespace flag as not provided', () => {
+    expect(resolveUploadContentType('photo.png', '')).toBe('image/png');
+    expect(resolveUploadContentType('photo.png', '   ')).toBe('image/png');
+  });
+
+  it('falls back to octet-stream for an unknown extension', () => {
+    expect(resolveUploadContentType('archive.unknownext')).toBe('application/octet-stream');
+    expect(resolveUploadContentType('noext')).toBe('application/octet-stream');
+  });
+});
--- a/src/commands/storage/upload.ts
+++ b/src/commands/storage/upload.ts
@@ -5,6 +5,7 @@ import { getProjectConfig } from '../../lib/config.js';
 import { requireAuth } from '../../lib/credentials.js';
 import { handleError, getRootOpts, CLIError, ProjectNotLinkedError } from '../../lib/errors.js';
 import { outputJson, outputSuccess } from '../../lib/output.js';
+import { mimeTypeFromName } from '../../lib/mime.js';
 
 export function registerStorageUploadCommand(storageCmd: Command): void {
   storageCmd
@@ -12,6 +13,7 @@ export function registerStorageUploadCommand(storageCmd: Command): void {
     .description('Upload a file to a storage bucket')
     .requiredOption('--bucket <name>', 'Target bucket name')
     .option('--key <objectKey>', 'Object key (defaults to filename)')
+    .option('--content-type <type>', 'MIME type to store (defaults to one inferred from the file extension)')
     .action(async (file: string, opts, cmd) => {
       const { json } = getRootOpts(cmd);
       try {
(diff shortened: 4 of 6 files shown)


jwfing (commented):
## Review: fix(storage): set content type on CLI uploads

**Summary:** A clean, well-scoped bug fix that infers the MIME type from the file extension (with a `--content-type` override) so CLI uploads stop defaulting to `application/octet-stream` — the root cause and fix are correctly identified.

**Requirements context:** No matching spec/plan found under `docs/specs/` (the only specs there cover the `diagnose` and `db-migrations` commands). Assessed against the PR description and Linear INS-358. The diff matches the stated intent precisely — three files, no scope creep.

---

### Critical
(none)

### Suggestion

**Software engineering — no test for the command-level resolution logic** (`src/commands/storage/upload.ts:36-43`)
The pure helper `mimeTypeFromName` is nicely covered, but the precedence chain that actually wires it up — explicit `--content-type` flag wins → inferred → `application/octet-stream` fallback, and that the resolved value lands on the `Blob` type — has no test. This repo does have command-level tests (e.g. `src/commands/projects/link.test.ts`, `src/commands/payments/transactions.test.ts`), so a small test asserting that `opts.contentType` overrides inference and that an unknown extension falls back would fit conventions and lock in the behavior. Non-blocking since the core logic is the well-tested pure helper.

**Functionality — inference runs on the full path, not the basename** (`src/lib/mime.ts:77-81`)
`mimeTypeFromName` calls `name.lastIndexOf('.')` on the whole string. For a path where a parent directory contains a dot and the file has no extension (e.g. `/home/user.name/datafile`), `ext` becomes `name/datafile`, which isn't a map key, so it returns `undefined`. This is *safe* — it can only ever degrade to the octet-stream fallback, never produce a wrong type, because no map key contains `/` — but it silently skips inference for those paths. Consider extracting the basename first (`basename` is already imported in `upload.ts`) so directory dots can't shadow a real extension.

### Information

- **`--content-type ""` (empty string)** (`src/commands/storage/upload.ts:36-37`): `??` only short-circuits on `null`/`undefined`, so an explicit empty-string flag value would pass through as `''`. A `Blob` with an empty type effectively serializes as octet-stream, so the outcome is benign, but `opts.contentType || mimeTypeFromName(file) || '...'` (or trimming) would treat empty as "not provided" if that's the desired semantics. Minor edge case.
- **Naming/imports** are consistent with the codebase (`.js` extension on relative imports, `src/lib/*.test.ts` colocated tests). The lookup table is a reasonable curated set; unknown extensions correctly fall back, so there's genuinely **no regression** as claimed.

---

### Dimension coverage

- **Software engineering:** Good — pure helper has thorough unit tests (case-insensitivity, multi-dot, missing/trailing-dot, dotfiles). Only gap is the untested command wiring (Suggestion above).
- **Functionality:** Solves INS-358 as described. Fallback chain is correct; the path-dot case (Suggestion) only ever degrades safely.
- **Security:** No security-relevant regressions. The `--content-type` user value flows into the `Blob` type → multipart `Content-Type`, but the `Blob` constructor normalizes the type per the WHATWG spec (characters outside U+0020–U+007E force the type to `""`), so CRLF/header injection is not possible. No secrets or auth handling touched; the `Authorization` header is unchanged.
- **Performance:** No concerns — a static O(1) lookup table; `readFileSync` is pre-existing and unchanged.

---

No blocking issues. The two Suggestions (a command-level test and basename extraction) and the Information notes are worth addressing but do not gate merge. Final GitHub approval remains a human action.

jwfing (dismissed):
LGTM - approved.

jwfing (approved):
LGTM - approved.

PR #191, opened 2026-07-07
feat(compute): add --always-on flag to opt out of scale-to-zero

tonychang04:
## What

Adds `--always-on` / `--scale-to-zero` flags to `compute deploy`. **Default unchanged** — omitting both flags sends no field: new services scale to zero, existing services keep their current setting.

- `--always-on` → `scaleToZero: false` in the create/update body (machine runs 24/7, no cold starts)
- `--scale-to-zero` → explicit `scaleToZero: true` to revert an always-on service on a later deploy
- Passing both is rejected as mutually exclusive
- Success output shows an `Always-on:` line when the service has scale-to-zero disabled

Wired into both deploy modes (image mode POST/PATCH and source mode prepare + final PATCH), same pattern as `--protocol`.

## Dependencies

- Backend counterpart: InsForge/InsForge#1638 (adds `scaleToZero` to the compute services API + Fly autostop mapping). Against an older backend the flag is silently ignored (zod strips unknown keys), and deploys without the flag are wire-identical to today.
- Cloud-managed mode passthrough: insforge-cloud-backend companion PR.
- `insforge-cli` agent-skill update (compute-deploy.md) ships as a separate insforge-skills PR.

## Validation

- `npm test` — 573 passed (4 new: body contains scaleToZero=false / =true, omitted when no flag, mutual-exclusion error)
- `npm run build` — clean
- `npx eslint` on changed files — clean
- `node dist/index.js compute deploy --help` shows the new flags

https://claude.ai/code/session_01MtvPScrRiH2k5RT7tNPXah

> [!NOTE]
> ### Add `--always-on` and `--scale-to-zero` flags to `compute deploy`
> - `--always-on` sets `scaleToZero=false` in the service create/update request; `--scale-to-zero` sets it to `true`; omitting both leaves the field unset
> - Passing both flags together is rejected with a mutually exclusive error
> - Prints a confirmation line when a service is deployed with scale-to-zero disabled
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ab4ce06.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added support for `compute deploy` scaling options to choose between always-on and scale-to-zero behavior.
  * The command now shows an “Always-on” message in non-JSON output when applicable.

* **Bug Fixes**
  * Prevents using `--always-on` and `--scale-to-zero` together by returning a clear validation error.
  * Preserves existing/default scaling behavior when neither flag is provided.

### DIFF
--- a/src/commands/compute/deploy.test.ts
+++ b/src/commands/compute/deploy.test.ts
@@ -70,3 +70,68 @@ describe('compute deploy --protocol', () => {
     ).rejects.toThrow(/Invalid --protocol/);
   });
 });
+
+describe('compute deploy --always-on / --scale-to-zero', () => {
+  beforeEach(() => {
+    ossFetchMock.mockReset();
+    ossFetchMock.mockResolvedValueOnce({ json: async () => [] }); // initial list
+    ossFetchMock.mockResolvedValueOnce({
+      json: async () => ({ name: 'api', status: 'running', endpointUrl: 'https://api.fly.dev', port: 8080, scaleToZero: false }),
+    });
+  });
+
+  it('includes scaleToZero=false in request body when --always-on', async () => {
+    const cmd = new Command();
+    cmd.exitOverride();
+    const compute = cmd.command('compute');
+    registerComputeDeployCommand(compute);
+    await cmd.parseAsync([
+      'node', 'lim', 'compute', 'deploy',
+      '--image', 'nginx', '--name', 'api', '--always-on',
+    ]);
+    const createCall = ossFetchMock.mock.calls[1];
+    const body = JSON.parse(createCall[1].body);
+    expect(body.scaleToZero).toBe(false);
+  });
+
+  it('includes scaleToZero=true in request body when --scale-to-zero (explicit revert)', async () => {
+    const cmd = new Command();
+    cmd.exitOverride();
+    const compute = cmd.command('compute');
+    registerComputeDeployCommand(compute);
+    await cmd.parseAsync([
+      'node', 'lim', 'compute', 'deploy',
+      '--image', 'nginx', '--name', 'api', '--scale-to-zero',
+    ]);
+    const createCall = ossFetchMock.mock.calls[1];
+    const body = JSON.parse(createCall[1].body);
+    expect(body.scaleToZero).toBe(true);
+  });
+
+  it('omits scaleToZero from body when neither flag is passed — create defaults server-side, update keeps the existing setting', async () => {
+    const cmd = new Command();
+    cmd.exitOverride();
+    const compute = cmd.command('compute');
+    registerComputeDeployCommand(compute);
+    await cmd.parseAsync([
+      'node', 'lim', 'compute', 'deploy',
+      '--image', 'nginx', '--name', 'api',
+    ]);
+    const createCall = ossFetchMock.mock.calls[1];
+    const body = JSON.parse(createCall[1].body);
+    expect('scaleToZero' in body).toBe(false);
+  });
+
+  it('rejects --always-on combined with --scale-to-zero', async () => {
+    const cmd = new Command();
+    cmd.exitOverride();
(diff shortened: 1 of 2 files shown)


jwfing (commented):
## Review: `feat(compute): add --always-on flag to opt out of scale-to-zero`

**Summary:** A tight, well-tested +87/-0 change that adds mutually-exclusive `--always-on` / `--scale-to-zero` flags to `compute deploy`, faithfully following the existing `--protocol` idiom with correct back-compat (omit both → no field sent).

**Requirements context:** No matching spec/plan found — `docs/specs/` contains only the diagnose-command and db-migrations designs, neither of which covers compute deploy flags. Assessed against the PR description and the surrounding code conventions alone.

---

### Findings

#### Critical
(none) — no correctness, security, data-loss, or back-compat-breaking issues found.

#### Suggestion

- **Functionality (UX) — no confirmation on the revert path** (`src/commands/compute/deploy.ts:185`, `:326`): the success output prints an `Always-on:` line only when `service.scaleToZero === false`. When a user passes `--scale-to-zero` to revert an always-on service back to scaling to zero, the response has `scaleToZero === true` and nothing is printed, so they get no confirmation the revert actually took effect. Consider a symmetric line for the re-enabled case (e.g. `Scale-to-zero: enabled`). Non-blocking.

- **Software engineering (test coverage) — only the image-mode POST path is exercised** (`src/commands/compute/deploy.test.ts:71-137`): all four new tests hit the create (POST) branch. The image-mode PATCH branch (which relies on `updateBody = { ...body }` at `deploy.ts:154` carrying `scaleToZero` through the spread) and the source-mode prepare+final-PATCH branch (`deploy.ts:229`, `:302`) are untested. The derivation logic lives in the shared create path so risk is low, and this matches the existing `--protocol` tests' coverage — but a single PATCH-path assertion would lock in that the flag survives an update of an existing service. Non-blocking.

#### Information

- **Nested ternary** (`src/commands/compute/deploy.ts:102-106`): the `alwaysOn ? false : scaleToZero ? true : undefined` chain is a nested ternary. It's clarified by the adjacent comment and the repo's eslint config has no `no-nested-ternary` rule, so this is purely a taste note — accept or ignore. The tri-state (`false` / `true` / `undefined`) it encodes is exactly right for the "send nothing when neither flag is set" contract.
- **Pre-existing, not introduced here:** the redundant `if (opts.protocol === 'tcp') updateBody.protocol = 'tcp';` at `deploy.ts:156` duplicates what the `{ ...body }` spread already carries. Out of scope for this PR; flagging only so it isn't mistaken for new code.

---

### Dimension coverage
- **Software engineering:** New tests are consistent with existing patterns (Command + `exitOverride` + `ossFetchMock` call-index assertions); import style, `CLIError` usage, and naming all match the file. See test-coverage suggestion above.
- **Functionality:** Correctly implements all four claimed behaviors — `--always-on`→`false`, `--scale-to-zero`→`true`, omit-both→field absent, both→rejected. Wired into all four request bodies (image POST/PATCH, source prepare/final PATCH). Back-compat is genuinely wire-identical to today when no flag is passed. Graceful against an older backend (missing `scaleToZero` in the response simply suppresses the output line). See revert-confirmation suggestion above.
- **Security:** No security-relevant changes — the new value is a derived boolean, never user-supplied string data reaching SQL/shell/HTTP; no secrets or PII logged; no auth/authz path touched; no new dependencies.
- **Performance:** No performance-relevant changes — no new network calls, loops, or per-request allocations; the field is folded into request bodies that were already being sent.

---

Zero Critical findings. The two Suggestions and Information notes are non-blocking; the author may address or defer them freely. Note this PR is marked **draft** and its behavior depends on backend PR InsForge/InsForge#1638 landing — worth confirming that ships (or is confirmed forward-compatible) before merge. Actual GitHub approval remains a separate human action.

jwfing (approved):
LGTM - approved.

PR #192, opened 2026-07-08
feat(diagnose): advisor to project OSS backend

CarmenDou:
## Summary by CodeRabbit

* **New Features**
  * Advisor diagnostics now prefer the OSS backend first, falling back to the legacy backend only when the advisor route is missing.
  * Advisor scan results can now be returned as “no scan available” without causing a failed diagnostic.

* **Bug Fixes**
  * Improved interpretation of advisor-related HTTP 404s, including clearer messaging for missing advisor support.
  * More consistent advisor diagnostic behavior across OSS and hosted environments.

> #### Changes since #192 opened
>
> - Added defensive null-checking to `fetchOssAdvisorIssues` function to normalize OSS backend responses [fee891f]
> 
>

> [!NOTE]
> 
> ## Summary by CodeRabbit
>
> ## Summary by CodeRabbit
>
> * **New Features**
>   * Advisor diagnostics now prefer the OSS backend first, falling back to the legacy backend only when the advisor route is missing.
>   * Advisor scan results can now be returned as “no scan available” without causing a failed diagnostic.
>
> * **Bug Fixes**
>   * Improved interpretation of advisor-related HTTP 404s, including clearer messaging for missing advisor support.
>   * More consistent advisor diagnostic behavior across OSS and hosted environments.
> 
> > #### Changes since #192 opened
> >
> > - Added defensive null-checking to `fetchOssAdvisorIssues` function to normalize OSS backend responses [fee891f]
> > - Added error handling for JSON parsing in `fetchOssAdvisorLatest` and `fetchOssAdvisorIssues` functions [a6efe05]
> 
> >
>

### DIFF
--- a/src/commands/diagnose/advisor.ts
+++ b/src/commands/diagnose/advisor.ts
@@ -48,14 +48,17 @@ function isOssAdvisorRouteMissing(err: unknown): boolean {
 
 async function fetchOssAdvisorLatest(): Promise<AdvisorScanSummary | null> {
   const res = await ossFetch('/api/advisor/latest');
-  return (await res.json()) as AdvisorScanSummary | null;
+  // Guard the parse: no-scan returns null and an empty body would make
+  // res.json() throw, matching the .catch pattern used across the codebase.
+  return (await res.json().catch(() => null)) as AdvisorScanSummary | null;
 }
 
 async function fetchOssAdvisorIssues(params: URLSearchParams): Promise<AdvisorIssuesResponse> {
   const res = await ossFetch(`/api/advisor/issues?${params.toString()}`);
   // The OSS backend returns { issues: [], total: 0 } for the no-scan state, but
-  // normalize defensively so a null/empty body can never crash the caller.
-  const data = (await res.json()) as AdvisorIssuesResponse | null;
+  // normalize defensively so a null/empty body can never crash the caller
+  // (res.json() throws on an empty body, so guard the parse itself too).
+  const data = (await res.json().catch(() => null)) as AdvisorIssuesResponse | null;
   return data ?? { issues: [], total: 0 };
 }
 
--- a/src/commands/diagnose/advisor.ts
+++ b/src/commands/diagnose/advisor.ts
@@ -53,7 +53,10 @@ async function fetchOssAdvisorLatest(): Promise<AdvisorScanSummary | null> {
 
 async function fetchOssAdvisorIssues(params: URLSearchParams): Promise<AdvisorIssuesResponse> {
   const res = await ossFetch(`/api/advisor/issues?${params.toString()}`);
-  return (await res.json()) as AdvisorIssuesResponse;
+  // The OSS backend returns { issues: [], total: 0 } for the no-scan state, but
+  // normalize defensively so a null/empty body can never crash the caller.
+  const data = (await res.json()) as AdvisorIssuesResponse | null;
+  return data ?? { issues: [], total: 0 };
 }
 
 async function fetchPlatformAdvisorLatest(
--- a/src/commands/diagnose/advisor.ts
+++ b/src/commands/diagnose/advisor.ts
@@ -1,8 +1,8 @@
 import type { Command } from 'commander';
-import { platformFetch } from '../../lib/api/platform.js';
+import { platformFetch, getProject } from '../../lib/api/platform.js';
 import { ossFetch } from '../../lib/api/oss.js';
 import { requireAuth } from '../../lib/credentials.js';
-import { handleError, getRootOpts, CLIError, ProjectNotLinkedError } from '../../lib/errors.js';
+import { handleError, getRootOpts, ProjectNotLinkedError } from '../../lib/errors.js';
 import { getProjectConfig, FAKE_PROJECT_ID } from '../../lib/config.js';
 import { outputJson, outputTable } from '../../lib/output.js';
 import { reportCliUsage } from '../../lib/skills.js';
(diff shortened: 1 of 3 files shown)


jwfing (commented):
## Summary

Routes `diagnose advisor` / `fetchAdvisorSummary` to each project's own OSS advisor first, falling back to the legacy cloud-backend Platform advisor only when the OSS route is missing (404) — enabling advisor in `--api-key`/OSS mode for the first time. The implementation is clean and the fallback model is sound, but the **PR description describes a different mechanism than the code actually ships**, and there's an asymmetric null-handling gap worth closing before merge.

## Requirements context

Closest spec: `docs/specs/2026-03-27-diagnose-command-design.md` (+ its implementation plan). It documents the *original* design, where OSS/`--api-key` mode **skips** advisor and advisor uses `platformFetch` (decision table row "OSS mode → skip metrics/advisor"; API-communication "advisor → platformFetch"). This PR deliberately supersedes those two decisions to consolidate advisor onto the project's OSS backend. **No spec/plan under `docs/specs/` (this repo has no `docs/superpowers/`) documents the OSS-routing migration itself** — so functionality is assessed against the PR description and the surrounding code.

## Findings

### Critical
(none)

### Suggestion

**1. Functionality — PR description describes version-gating; the code implements 404-based fallback.** `advisor.ts:38-113`, `oss.ts:216-223`
The title ("version-gate … >= 2.2.7"), the **What** section (read `/api/health` `{version}`, compare with `parseSemver`/`isVersionGte`, `>= 2.2.7` → OSS else cloud), and the **Verification** note ("the new semver helpers `parseSemver`/`isVersionGte` are currently uncovered") all describe an implementation that isn't in the diff. A grep of `src/` finds no `parseSemver`, `isVersionGte`, or `/api/health` call. What actually ships is *try-OSS-first, fall back to Platform on a route-level 404* (`isOssAdvisorRouteMissing`, `advisor.ts:44-46`). The 404 approach is arguably better (no extra round-trip, no trusting stale metadata), so this is a description bug, not a code bug — but the derived migration notes are inaccurate (there is no numeric `2.2.7` gate; `--api-key` mode on an old backend surfaces the generic `oss.ts:222` "Backend Advisor is not available … upgrade" message rather than a "`>= 2.2.7`" one). Please rewrite the description to match the shipped 404-fallback design before merge so reviewers and the changelog aren't misled.

**2. Functionality — asymmetric null-handling on the OSS issues path may crash the new primary flow.** `advisor.ts` (`fetchOssAdvisorIssues`, lines ~50-53; command body ~142-160)
`fetchOssAdvisorLatest` correctly tolerates a null body (`AdvisorScanSummary | null`) and the command guards it (`if (!scan) console.log('No scan yet.')`). But `fetchOssAdvisorIssues` casts straight to `AdvisorIssuesResponse` with no null guard. The `oss.ts:217-219` comment asserts the OSS advisor endpoints "return **200 with a null/empty body** when no scan exists" — that is exactly the no-scan state the null-scan guard anticipates. If `/api/advisor/issues` returns a null body in that state, then both branches dereference null:
- table: `if (!issuesData.issues || issuesData.issues.length === 0)` → `TypeError` on null `issuesData`
- json: `outputJson({ scan, issues: issuesData.issues })` → `TypeError`

This would crash the brand-new "OSS backend, route present, no scan yet" path — the headline use case of this PR. The Platform equivalent (`fetchPlatformAdvisorIssues`) always returns `{ issues: [], total: 0 }`, so it's safe. Recommend normalizing in `fetchOssAdvisorIssues` (`return (await res.json()) ?? { issues: [], total: 0 }`). **This would be Critical if InsForge #1649's `/api/advisor/issues` returns null/empty for the no-scan state — please verify against that PR; the fix is cheap regardless.**

**3. Software engineering — no test coverage for the new routing/fallback logic.** `advisor.ts`, `index.ts`
`find` shows no test file for diagnose/advisor, and the PR adds real branching — OSS-first, route-missing→Platform fallback, `ossMode` no-fallback, null-scan handling, the new `isOssAdvisorRouteMissing` helper — with zero tests (the PR acknowledges this). Per the repo's `insforge-development` / TDD convention, each new behavior should have a test, and the finding above (#2) is precisely the close-call a unit test would surface. Suggest covering: OSS success, OSS 404 → Platform fallback, `ossMode` rethrow (no fallback), null scan → "No scan yet.", and null/empty issues.

**4. Software engineering — `isOssAdvisorRouteMissing` is looser than the `oss.ts` route-level guard.** `advisor.ts:44-46` vs `oss.ts:191,221`
`oss.ts` only rewrites the advisor 404 message when `isRouteLevel404` (`!err.error || err.error === 'NOT_FOUND'`), deliberately distinguishing a route miss from a resource miss carrying a structured error code. `isOssAdvisorRouteMissing` classifies **any** `statusCode === 404` as route-missing and triggers the cloud fallback. It works today because these helpers only ever hit `/api/advisor/*` and the documented contract says advisor never resource-404s — but if that contract ever slips (a resource-level 404 with an error code), the CLI would silently fall back to cloud-backend and show stale data instead of erroring. Consider mirroring `oss.ts`'s check via the `CLIError` error code (3rd constructor arg) for defensiveness.

### Information

**5. Functionality — redundant double-fetch in the fallback catch.** `advisor.ts` command body (~150-159)
If `fetchOssAdvisorLatest()` succeeds but `fetchOssAdvisorIssues()` then throws a route-missing 404, the catch re-fetches **both** scan and issues from Platform, discarding the good OSS scan. In practice `/latest` and `/issues` ship together so this shouldn't occur; noting only for completeness.

**6. Security — no concerns.** No new user input reaches SQL/shell; `severity`/`category`/`limit` are passed through `URLSearchParams` (encoded). No secrets newly logged (`platform.ts` redacts the bearer). Enabling advisor in `--api-key` mode doesn't weaken authz — the api_key already grants superadmin OSS reads; advisor is just another OSS read.

**7. Performance — no concerns.** Advisor runs inside the existing `Promise.allSettled` fan-out; OSS mode now issues one advisor request where it previously short-circuited — expected cost of the feature. No N+1, hot-loop, or blocking work introduced.

## Verdict

**approved** (informational — a human still gives the GitHub approval). No Critical findings. The routing/fallback design is sound; before merge please (a) fix the PR description to match the 404-fallback implementation, and (b) verify/guard the null-issues path in #2 (cheap, and Critical if the OSS issues endpoint returns null for no-scan). Adding the missing unit tests would catch exactly that case.

jwfing (dismissed):
LGTM - approved.

Fermionic-Lyu (approved):
LGTM, Approved.

PR #201, opened 2026-07-22
fix(branch): make `branch create` success mean the branch is usable

iPLAYCAFE-dev:
Follow-up to [InsForge/InsForge#1790](https://github.com/InsForge/InsForge/issues/1790), fixing the half of it that lives in this repo. The root cause of the readiness gap is cloud-side, but the CLI is where it becomes a broken user experience, and all three of these are fixable here.

Found while automating branch creation on `ap-southeast`. Every measurement below is from real branches, all of which were deleted afterwards.

## 1. `ready` is a control-plane state, not readiness

`pollUntilReady` returns the moment `branch_state === 'ready'` and never contacts the branch's own host. But `branch_state` flips when the provisioning job returns, while the instance is still coming up — every request to `https://<appkey>.<region>.insforge.app` resets until it does.

**Measured: the host started serving at t+2m in one case and t+11.5m in another, with `branch_state` reading `ready` the whole time.**

Because `create` then auto-switches the directory onto that host, the failure surfaces on the user's *next* command rather than here:

```
$ insforge db query "select 1"
{"error":"fetch failed","code":"UNKNOWN_ERROR"}
```

Now: once the control plane says ready, poll `GET /api/health` on the branch itself until it answers, and say so in the spinner. If it never answers within the budget, the command reports that honestly instead of claiming success — the branch's name and id are still printed, because it is real and it is billing.

## 2. The 5-minute poll ceiling was below the observed provisioning time

`POLL_TIMEOUT_MS` was 5 minutes, so the 11.5-minute branch was reported as *"still in 'creating' state"* when it was simply not finished yet. Raised to 15, with a separate 10-minute budget for the data-plane wait.

## 3. A failed `create` can leave a live branch behind

`createBranchApi` carries no idempotency key. A transport failure on the **response** leg — the POST arrived and the branch was created — throws before `created` is bound, so the CLI exits non-zero with no id and no name while a branch exists and bills. We hit exactly this:

```
$ insforge branch create <name> --mode schema-only --no-switch
{"error":"Connection to api.insforge.dev was reset. A proxy, VPN, or firewall may be interfering."}
# exit 1
$ insforge branch list
# the branch is there, branch_state "creating"
```

It then refused `delete` for about five minutes with `"Branch is currently busy (creating or merging)."` before the state flipped and the delete succeeded.

Now: on a create failure, ask `branch list` — a control-plane call, so it still works while the branch's own host is unreachable — whether a branch exists under that name, and adopt it if so. The original error is rethrown unchanged when nothing was created.

## Also

`ossFetch` called `fetch` unguarded, unlike `platformFetch`, so a dead data plane surfaced as the generic `UNKNOWN_ERROR` instead of naming the host. The new `probeBackendHealth` routes its errors through `formatFetchError`, so `Connection to <host> was reset` is what a caller sees. That one line cost us a lot of triage time.

## Notes for review

- **New tests** cover the not-serving path, the adopt path, and that a genuine failure with nothing created still exits non-zero.
- **The existing create tests needed a mock** for the new probe: unmocked, it makes a real request to a fake host and then polls. `beforeEach` resets it explicitly, because `clearAllMocks` keeps implementations and an unreachable branch would otherwise leak into every later test.
- `npm run build` passes; `vitest` passes for everything I touched. Four unrelated failures (`flyctl` ×3, `migrations` ×1) reproduce on a clean `main` on this machine (Windows) and are untouched by this change.
- **Behaviour change worth calling out:** a branch that reports `ready` but never serves now stops the spinner with an error frame instead of a success line. That seemed right — it is the case where continuing would break the user's next command — but say the word if you would rather it warn and exit 0.

## Agent skills

`insforge-cli/references/branch/overview.md` says a branch takes 30–120s and that `ready` means *"usable — can be switched, modified, merged, or reset"*, and its post-create checklist goes straight to `functions deploy`. Both look worth updating in `InsForge/agent-skills` once you have confirmed the timing on your side — happy to open that PR alongside this one.

## Open question this does not answer

Is there a branch-scoped signal that reflects the branch's own data plane rather than the provisioning job — or would you accept `waitForCompletion=true` on `POST /projects/v1/{id}/branches`, matching the convention already used by `projects restart` and `backups create`? `/api/health` is the best thing available to a client today, but it is a static handler that never touches Postgres, so it proves the process is listening rather than that the branch is usable. If a better signal exists or is planned, this polling should key off that instead.

---
## Summary by cubic
Make `branch create` report success only when the branch is actually usable. It now waits for the branch host to serve, fails on stuck provisioning, and adopts an existing branch only in safe, verified cases after transport errors.

- **Bug Fixes**
  - Wait for the data plane: after control‑plane `ready`, poll the branch host’s `/api/health`; include `serving` in `--json`, print the branch identity first, and exit non‑zero if not serving within 10 min.
  - Honest exit on stuck provisioning: exit non‑zero if the branch never reaches `ready` within the 15‑minute poll budget.
  - Longer budgets: control‑plane poll raised to 15 min; health probe every 5s with a 10‑min cap.
  - Safe adoption on transport failures: only adopt after a tagged network error (`NETWORK_ERROR_CODE`), only if a same‑name branch was created at/after the request time (60s skew) and with a matching mode. Never adopt on API rejections (e.g., duplicate name) or pre‑existing branches.
  - Clearer connectivity errors: new `probeBackendHealth` routes through `formatFetchError`, so messages name the branch host.

<sup>Written for commit ea18b8122e2897b91cd6dd7a660a65fe4705902c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/CLI/pull/201?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `branch create` by adopting a matching existing branch when failures are due to transient connection issues.
  * Added stricter readiness checks: the CLI now confirms the branch host is actually reachable before reporting success or switching.
  * Updated outcomes and messaging when the branch is “ready” but not yet serving, including correct non-zero exit behavior for timeout/failed provisioning.
  * Preserved original failures when adoption isn’t applicable (e.g., duplicate-name rejections or no existing branch found).
* **Tests**
  * Expanded coverage for edge cases around provisioning/adoption and serving reachability timeouts.

> [!NOTE]
> ### Fix `branch create` to only report success when the branch is ready and serving
> - Adds a `waitUntilServing` step after provisioning completes: polls `/api/health` on the branch host every 5 seconds for up to 10 minutes before reporting success.
> - Increases the provisioning poll timeout from 5 to 15 minutes and exits non-zero if the branch never becomes ready or never starts serving.
> - Introduces `createBranchOrAdopt`: on a transport-layer failure during creation, lists branches and adopts a matching one (by name, mode, and `created_at`) instead of failing immediately.
> - Tags fetch-level errors with `NETWORK_ERROR_CODE` in `platformFetch` so callers can distinguish transport failures from HTTP errors.
> - In JSON mode, emits branch identity before raising any failure so callers still get the branch reference on non-zero exit.
> - Behavioral Change: `branch create` now exits non-zero if the branch provisions but never serves within the health timeout, where previously it would exit zero after provisioning.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized ea18b81.</sup>
> 
>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@insforge/cli",
-  "version": "0.1.99",
+  "version": "0.2.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@insforge/cli",
-      "version": "0.1.99",
+      "version": "0.2.0",
       "license": "Apache-2.0",
       "dependencies": {
         "@clack/prompts": "^0.9.1",
--- a/src/commands/branch/create.test.ts
+++ b/src/commands/branch/create.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect, vi, beforeEach, type Mock } from 'vitest';
 import { Command } from 'commander';
 import { registerBranchCreateCommand } from './create.js';
+import { CLIError } from '../../lib/errors.js';
 
 vi.mock('../../lib/api/platform.js', () => ({
   createBranchApi: vi.fn(async (_parentId: string, body: { mode: string; name: string }) => ({
@@ -26,6 +27,7 @@ vi.mock('../../lib/api/platform.js', () => ({
     branch_metadata: { mode: 'full' },
   })),
   listBranchesApi: vi.fn(async () => []),
+  NETWORK_ERROR_CODE: 'NETWORK_ERROR',
 }));
 
 // The data-plane readiness probe. It MUST be mocked: unmocked it makes a real
@@ -334,7 +336,7 @@ describe('branch create', () => {
     });
     const { createBranchApi, listBranchesApi } = await import('../../lib/api/platform.js');
     (createBranchApi as Mock).mockRejectedValueOnce(
-      new Error('Connection to api.insforge.dev was reset.'),
+      new CLIError('Connection to api.insforge.dev was reset.', 1, 'NETWORK_ERROR'),
     );
     (listBranchesApi as Mock).mockResolvedValueOnce([
       {
@@ -368,7 +370,9 @@ describe('branch create', () => {
       org_id: 'o1',
     });
     const { createBranchApi, listBranchesApi } = await import('../../lib/api/platform.js');
-    (createBranchApi as Mock).mockRejectedValueOnce(new Error('boom'));
+    (createBranchApi as Mock).mockRejectedValueOnce(
+      new CLIError('Connection to api.insforge.dev was reset.', 1, 'NETWORK_ERROR'),
+    );
     (listBranchesApi as Mock).mockResolvedValueOnce([]);
     const program = new Command().exitOverride();
     program.option('--json').option('--api-url <url>').option('-y, --yes');
@@ -389,6 +393,137 @@ describe('branch create', () => {
       process.exit = origExit;
       process.stderr.write = origStderr;
     }
+    expect(exitCode).toBe(1);
+  });
+  it('does NOT adopt on an API rejection — a duplicate name is a refusal, not a lost response', async () => {
+    // Adopting here would switch the caller into a pre-existing branch with a
+    // different mode and different data. Only a transport failure is ambiguous.
+    const { getProjectConfig } = await import('../../lib/config.js');
+    (getProjectConfig as Mock).mockReturnValue({
+      project_id: 'p1',
+      project_name: 'parent',
+      org_id: 'o1',
+    });
+    const { createBranchApi, listBranchesApi } = await import('../../lib/api/platform.js');
+    (createBranchApi as Mock).mockRejectedValueOnce(
+      new CLIError("Branch name 'feat-x' already exists on this parent", 1),
+    );
+    const program = new Command().exitOverride();
+    program.option('--json').option('--api-url <url>').option('-y, --yes');
+    registerBranchCreateCommand(program);
+    let exitCode: number | undefined;
+    const origExit = process.exit;
+    process.exit = ((code?: number) => {
+      exitCode = code;
+      throw new Error('__exit__');
+    }) as typeof process.exit;
+    const origStderr = process.stderr.write.bind(process.stderr);
+    process.stderr.write = (() => true) as typeof process.stderr.write;
+    try {
+      await program
+        .parseAsync(['create', 'feat-x', '--mode', 'schema-only', '--no-switch'], { from: 'user' })
+        .catch(() => {});
+    } finally {
+      process.exit = origExit;
+      process.stderr.write = origStderr;
+    }
+    expect(listBranchesApi as Mock).not.toHaveBeenCalled();
+    expect(exitCode).toBe(1);
+  });
+
+  it('does NOT adopt a branch that predates the request', async () => {
+    // Same name, but it existed before we asked — so it is not ours.
+    const { getProjectConfig } = await import('../../lib/config.js');
+    (getProjectConfig as Mock).mockReturnValue({
+      project_id: 'p1',
+      project_name: 'parent',
+      org_id: 'o1',
+    });
+    const { createBranchApi, listBranchesApi } = await import('../../lib/api/platform.js');
+    (createBranchApi as Mock).mockRejectedValueOnce(
+      new CLIError('Connection to api.insforge.dev was reset.', 1, 'NETWORK_ERROR'),
+    );
+    (listBranchesApi as Mock).mockResolvedValueOnce([
+      {
+        id: 'someone-elses',
+        parent_project_id: 'p1',
+        organization_id: 'o1',
+        name: 'feat-x',
+        appkey: 'p1ky-old',
+        region: 'us-east',
+        branch_state: 'ready',
+        branch_created_at: new Date(Date.now() - 48 * 60 * 60 * 1000).toISOString(),
+        branch_metadata: { mode: 'full' },
+      },
+    ]);
+    const program = new Command().exitOverride();
+    program.option('--json').option('--api-url <url>').option('-y, --yes');
+    registerBranchCreateCommand(program);
+    let exitCode: number | undefined;
+    const origExit = process.exit;
+    process.exit = ((code?: number) => {
+      exitCode = code;
+      throw new Error('__exit__');
+    }) as typeof process.exit;
+    const origStderr = process.stderr.write.bind(process.stderr);
+    process.stderr.write = (() => true) as typeof process.stderr.write;
+    try {
+      await program
+        .parseAsync(['create', 'feat-x', '--mode', 'schema-only', '--no-switch'], { from: 'user' })
+        .catch(() => {});
+    } finally {
+      process.exit = origExit;
+      process.stderr.write = origStderr;
+    }
+    expect(exitCode).toBe(1);
+  });
+
+  it('exits non-zero when the branch never serves, but still emits its identity first', async () => {
+    // The branch exists and is billing: automation must be able to find and
+    // delete it even though the command is failing.
+    const { getProjectConfig } = await import('../../lib/config.js');
+    (getProjectConfig as Mock).mockReturnValue({
+      project_id: 'p1',
+      project_name: 'parent',
+      org_id: 'o1',
+    });
+    const { probeBackendHealth } = await import('../../lib/api/oss.js');
+    (probeBackendHealth as Mock).mockResolvedValue({ reachable: false, status: null });
+    const lines: string[] = [];
+    const origLog = console.log;
+    console.log = ((...args: unknown[]) => {
+      lines.push(args.join(' '));
+    }) as typeof console.log;
+    const origStderr = process.stderr.write.bind(process.stderr);
+    process.stderr.write = (() => true) as typeof process.stderr.write;
+    let exitCode: number | undefined;
+    const origExit = process.exit;
+    process.exit = ((code?: number) => {
+      exitCode = code;
+      throw new Error('__exit__');
+    }) as typeof process.exit;
+    vi.useFakeTimers();
+    const program = new Command().exitOverride();
+    program.option('--json').option('--api-url <url>').option('-y, --yes');
+    registerBranchCreateCommand(program);
+    try {
+      const run = program
+        .parseAsync(['create', 'feat-x', '--mode', 'schema-only', '--no-switch', '--json'], {
+          from: 'user',
+        })
+        .catch(() => {});
+      await vi.runAllTimersAsync();
+      await run;
+    } finally {
+      vi.useRealTimers();
+      console.log = origLog;
+      process.exit = origExit;
+      process.stderr.write = origStderr;
+    }
+    const payload = lines.join('\n');
+    expect(payload).toContain('branch-id');
+    expect(payload).toContain('"serving"');
+    expect(payload).toContain('false');
     expect(exitCode).toBe(1);
   });
 });
--- a/src/commands/branch/create.test.ts
+++ b/src/commands/branch/create.test.ts
@@ -70,6 +70,27 @@ vi.mock('@clack/prompts', () => ({
   spinner: () => spinnerMock,
 }));
 
+// Run `fn` with process.exit + stderr captured, and always restore them. Returns
+// the exit code fn triggered (undefined if it never exited). Timer/mock lifecycle
+// stays with the caller — this only owns the exit/stderr swap.
+async function withCapturedExit(fn: () => Promise<void>): Promise<number | undefined> {
+  let exitCode: number | undefined;
+  const origExit = process.exit;
+  const origStderr = process.stderr.write.bind(process.stderr);
+  process.exit = ((code?: number) => {
+    exitCode = code;
+    throw new Error('__exit__');
+  }) as typeof process.exit;
+  process.stderr.write = (() => true) as typeof process.stderr.write;
+  try {
+    await fn();
+  } finally {
+    process.exit = origExit;
+    process.stderr.write = origStderr;
+  }
+  return exitCode;
+}
+
 describe('branch create', () => {
   beforeEach(async () => {
     vi.clearAllMocks();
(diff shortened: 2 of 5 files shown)


iPLAYCAFE-dev:
Closes #202.

@agent-zhang-beihai — apologies, I submitted before reading the contribution workflow. Issue #202 is open now with the measurements and repro, and I've asked for it to be assigned. Happy to close this PR and resubmit after assignment if you'd prefer the order kept strictly.

---

**Both review findings were right, and both are fixed in `552cf50`.** Thanks — the second one in particular is the same defect this PR exists to remove, reproduced inside the fix for it.

### 1. Adoption was too broad — @greptile-apps (P1/security), @cubic-dev-ai (P1)

A duplicate-name response is a **refusal**, not a lost response, and adopting on it could switch the caller into a pre-existing branch with a different mode and different data. Two guards now, either sufficient on its own:

- **Only a tagged transport failure is eligible.** `platformFetch`'s fetch-level catch now throws `CLIError` with code `NETWORK_ERROR`, so "the server said no" and "we never heard back" are finally distinguishable by callers — previously both were a bare `CLIError`, which is why I could not narrow this correctly the first time. Every HTTP/API rejection rethrows untouched, and the test asserts `listBranchesApi` is not even *called* in that case.
- **The candidate must post-date the request.** `branch_created_at >= requestedAt`, so a pre-existing same-name branch can never be adopted regardless of how the error is classified.

On the skew allowance: I used 60s. Being slightly wide risks adopting a branch someone else created seconds ago under the same name; being too narrow re-opens the orphaned-billing bug. If you'd rather have zero tolerance and accept the clock-skew false negative, say so and I'll drop it — it's a one-line change and your call, since you know how the control plane stamps `branch_created_at`.

### 2. A branch that never serves exited 0 — @greptile-apps (P1), @cubic-dev-ai (P1)

Correct, and it was worse than the comments say: I stopped the spinner with an error frame and left it there, so the process still exited 0, `--json` still emitted a ready-looking branch with no serving field, and the non-JSON path still advised re-sourcing the env.

Now `serving` is tracked separately from `provisioned`, `--json` emits `{ branch, serving }`, the env hint only prints when the host actually answers, and the command throws afterwards so the exit is non-zero.

**The identity is emitted before the throw on purpose:** the branch exists and is billing, so a caller must be able to find and delete it even as the command fails. That's also why the not-serving path prints the `branch delete` command rather than just an error.

### Verification

- 12 tests in `create.test.ts`, all passing — including the three new ones (API rejection not adopted, pre-existing branch not adopted, never-serving exits 1 while still printing the id and `serving:false`).
- `npm run build` passes; `tsc --noEmit` reports nothing for the three files I touched.
- Four unrelated failures (`flyctl` ×3, `migrations` ×1) reproduce on a clean `main` on this machine (Windows) and are untouched.

### Still open, and it's a maintainer question

`/api/health` is the best signal available to a client today, but it's a static handler that never touches Postgres — it proves the process is listening, not that the branch is usable. If there's a branch-scoped readiness signal, or you'd accept `waitForCompletion=true` on `POST /projects/v1/{id}/branches` (matching `projects restart` / `backups create`), this polling should key off that instead. Asked on InsForge/InsForge#1790 as well.

iPLAYCAFE-dev:
Thanks for the re-review. Both threads resolved on cubic as well, and all four checks are green on `552cf50`.

**On the stamping semantics — I have data for this, and it settles the skew question.**

`branch_created_at` is stamped at **creation**, not at job completion. Two observations from the spike, both from raw `branch create` / `branch list` payloads:

- `branch create` returns `branch_created_at` **in its own response**, alongside `branch_state: "creating"` — so the row is stamped before provisioning has done anything.
- One branch was created at `2026-07-22T05:08:25.682Z` and was still `branch_state: "creating"` when listed at `05:13` (its `updated_at` had moved to `05:13:07.124Z`). `branch_created_at` stayed at `05:08:25.682Z` throughout, and `created_at === branch_created_at` on that row.

So the loaded-control-plane case you were worried about doesn't apply: the stamp can't drift later than the actual creation, and it lands within the same request. That leaves only genuine clock skew between the client and the control plane, which is what the 60s window is for. I'll leave it at 60s unless a maintainer wants it tighter.

**On the missing final probe after `HEALTH_TIMEOUT_MS`** — agreed on both counts: the gap is one 5s interval against a 10-minute budget, and the failure direction is safe (we report *not serving* for a branch that just became reachable, and the branch's identity is still emitted so nothing is orphaned). I considered adding a probe after the loop for an exact postcondition and decided against it — you've already LGTM'd this commit and the change would be churn for a sub-1% timing window. Happy to add it if a maintainer would rather have the loop's postcondition be exact.

Awaiting human review. Tracking issue is #202 as requested — and to be explicit about what this PR does and doesn't do: it stops the CLI *reporting* readiness it hasn't verified, but the underlying gap (a branch reporting `ready` minutes before its host serves) is cloud-side, and the question of whether a branch-scoped readiness signal exists — or whether `waitForCompletion=true` on `POST /projects/v1/{id}/branches` would be accepted, matching `projects restart` / `backups create` — is still open on InsForge/InsForge#1790. If either lands, this polling should key off it instead of `/api/health`, which only proves the process is listening.

iPLAYCAFE-dev:
Process box ticked: #202 is now assigned to me (thanks @agent-zhang-beihai), and this PR closes it — so the issue-first workflow is satisfied for anyone reviewing later.

Current state, so a human reviewer doesn't have to reconstruct it:

- **All four checks green** on `552cf50` — CodeRabbit, Greptile, cubic, Macroscope (skipped).
- **Both P1 findings fixed and re-reviewed.** cubic marks each "✅ Addressed"; Greptile: *"Both threads resolved. LGTM on 552cf50."*
- **Greptile's follow-up question is answered with data** in the comment above: `branch_created_at` is stamped at **creation**, not at job completion — `branch create` returns it in its own response alongside `branch_state: "creating"`, and on a branch that stayed `creating` for five minutes the value never moved. So the skew window only has to absorb client↔control-plane clock drift, which is what the 60s is for.

Two things still open that only a maintainer can decide, both already stated above and neither blocking:

1. whether to tighten or drop the 60s `branch_created_at` skew allowance;
2. whether a branch-scoped readiness signal exists or `waitForCompletion=true` on `POST /projects/v1/{id}/branches` would be accepted — if so this should poll that instead of `/api/health`, which only proves the process is listening.

Nothing further from me unless review turns something up.

jwfing (commented):
## Review — `fix(branch): make `branch create` success mean the branch is usable`

**Summary:** A well-scoped, well-tested fix that makes `branch create` gate success on data-plane readiness, safely adopts a branch orphaned by a transport failure, and raises the poll budget to match observed provisioning times — no blocking issues found.

### Requirements context
No matching spec/plan found under `docs/specs/` (the only specs there cover `diagnose` and `db-migrations`; this repo has no `docs/superpowers/`). Assessed against the PR description and the linked issue [InsForge/InsForge#1790](https://github.com/InsForge/InsForge/issues/1790). The three stated problems (ready ≠ serving, sub-provisioning poll ceiling, orphaned branch on response-leg failure) are each addressed, and the implementation choices match the description faithfully.

### Findings

#### Critical
(none)

#### Suggestion

**Functionality — inconsistent exit semantics for the other "not usable" outcome.** `src/commands/branch/create.ts:158-164` now correctly exits non-zero when a branch is `ready` but never serves. But the sibling path — `pollUntilReady` times out with the branch still in a non-terminal `creating` state (`create.ts:149-153`) — still emits only an info line and exits **0**. Given this PR's explicit goal ("success means usable"), a branch stuck provisioning past the 15-minute budget is equally unusable, yet automation reading the exit code sees success. Consider making that path exit non-zero too (or documenting why the two are treated differently). Low blast radius, hence not blocking.

**Functionality — `/api/health` proves "listening", not "usable".** `waitUntilServing` (`create.ts:228-245`) treats any `res.ok` from `probeBackendHealth` as serving. As you already note in the PR's open question, `/api/health` is a static handler that never touches Postgres, so a branch can pass this probe before its DB/schema is actually ready — a narrower version of the very gap this PR closes. This is the best client-side signal available today and you've flagged it honestly; noting it so it isn't lost. If `waitForCompletion=true` (matching `projects restart` / `backups create`) or a data-plane-scoped signal lands cloud-side, this poll should key off that instead.

#### Information

**Software engineering — `export const` interleaved in the import block.** `src/lib/api/platform.ts:4-6` places `export const NETWORK_ERROR_CODE` between two `import` statements. It's legal (imports hoist) and the lint gate passes (no `import/first` rule in `eslint.config.js`), but conventionally declarations sit below the import block — worth relocating for readability.

**Software engineering — double error surface on the not-serving path (non-JSON).** In the ready-but-not-serving case the spinner stops with an error frame (`create.ts:114-118`) and then the thrown `CLIError` also prints `Error: …` via `handleError`. Two messages for one condition; harmless and arguably informative, just noting.

### Notes on the good parts
- **Adoption guards are exactly right.** `createBranchOrAdopt` (`create.ts:197-219`) adopts only on a tagged `NETWORK_ERROR_CODE` transport failure *and* only for a same-name branch created at/after `requestedAt` (with a deliberate 60s skew), and rethrows the original error when `listBranchesApi` finds nothing or itself fails (`.catch(() => undefined)`). API rejections (duplicate name/quota/auth) correctly fall through untouched. The `NETWORK_ERROR_CODE` tagging in `platform.ts:101-115` is the clean way to distinguish transport failure from an HTTP response.
- **Test coverage matches the behavior changes** — not-serving exit, adopt-on-transport-failure, rethrow-when-nothing-created, no-adopt-on-API-rejection, no-adopt-on-predating-branch, and identity-emitted-before-failure are all exercised. The `beforeEach` re-stub of `probeBackendHealth` (test file) with the explanatory comment about `clearAllMocks` keeping implementations is a good catch that prevents cross-test leakage.
- **Security:** no new user input reaches SQL/shell; `probeBackendHealth` fetches an unauthenticated URL built from server-assigned `appkey`/`region` with a bounded 10s `AbortSignal.timeout`; no secrets logged (errors routed through `formatFetchError`, which emits host + code only). No auth/authorization changes.
- **Performance:** both poll loops are sequential and bounded (health: 5s × up to 10 min; provisioning: 3s × up to 15 min) — appropriate for a CLI, no hot-path concerns. `package-lock.json` change is only the 0.2.0 version sync (matches `package.json`), so no `npm ci` breakage.

### Verdict
**approved** (informational — no Critical findings; the two Suggestions and Information notes are non-blocking). Human green-check remains a separate action.

iPLAYCAFE-dev:
Thanks for the thorough review — all four points addressed in 42f9be6.

**Suggestion — inconsistent exit semantics.** Fixed. The stuck-provisioning path (`create.ts`, branch never reaches `'ready'` within `POLL_TIMEOUT_MS`) now throws a `CLIError` and exits non-zero, matching the ready-but-not-serving path. You're right that "success means usable" has to cover both unusable outcomes; a branch still `creating` past the 15-min budget is as unusable as one that's `ready` but never serves. Added a test (`getBranchApi` stuck in `'creating'` → exit 1), restoring the shared mock implementation in a `finally` so the `'creating'` stub can't leak the full poll budget into later tests.

**Suggestion — `/api/health` proves "listening", not "usable".** Acknowledged and left as the PR's documented open question. It's the best client-side signal available today; if `waitForCompletion=true` or a data-plane-scoped readiness signal lands cloud-side, `waitUntilServing` should key off that instead. Noted so it isn't lost.

**Information — `export const` interleaved in the import block.** Fixed — `NETWORK_ERROR_CODE` now sits below the full import block in `platform.ts`.

**Information — double error surface on the not-serving path.** Left as-is intentionally: the spinner error frame names the branch state and the `CLIError` gives the actionable "did not start serving within N minutes" line, so the two messages carry different information. Happy to collapse them if you'd prefer a single surface.

Build + lint clean; the new test and the existing branch-create suite pass. (The 4 unrelated `flyctl`/`migrations` failures are pre-existing Windows path-separator issues — confirmed they fail identically without this change — and pass on Linux CI.)

iPLAYCAFE-dev:
Both follow-up reviews addressed in `ea18b81`.

**cubic (P2) — collaborator same-name adoption in the skew window.** Valid, fixed. `createBranchOrAdopt` now also requires the candidate's **mode** to match the requested mode, so adoption needs same name AND same mode AND creation inside the window AND our own tagged transport failure — a much more specific coincidence than name alone. Test added: a same-name, *different*-mode branch created in the window is **not** adopted (the transport error propagates → non-zero exit). On the "use the actual send time" half: I've kept the 60s backward skew, for the reason stated earlier in the thread — `branch_created_at` is stamped at creation and never drifts, so the window only has to absorb client↔control-plane clock skew; dropping it risks excluding our *own* legitimately-created branch. The mode match closes the collision cubic is pointing at without reintroducing that clock-skew fragility. The real fix remains a server-issued idempotency/request token on `createBranchApi` (flagged in InsForge/InsForge#1790); this is the tightest client-side guard until that exists.

**CodeRabbit (nitpick) — repeated `process.exit`/`stderr` boilerplate.** Done. Extracted `withCapturedExit(fn)` and applied it to the tests this PR added (including the fake-timer provisioning-timeout one, where the timer/mock lifecycle stays with the caller and the helper just owns the exit/stderr swap). I left the three pre-existing blocks (`rejects when no project linked`, invalid-mode, switch-failure) as-is to keep this diff scoped to the PR's surface — happy to sweep them in the same pass if you'd prefer full consistency.

Build + lint clean; `create.test.ts` is 14/14. (The unrelated `flyctl`/`migrations` failures remain pre-existing Windows path-separator issues that pass on Linux CI.)

iPLAYCAFE-dev:
@jwfing — re-review request (I don't have permission to re-request formally from a fork, so flagging here). Two commits landed **after** your review:

- `42f9be6` — your review's two Suggestions: `branch create` now exits non-zero on a stuck-provisioning timeout too (matching the not-serving exit), and the `NETWORK_ERROR_CODE` export moved below the import block.
- `ea18b81` — cubic's follow-up P2: `createBranchOrAdopt` now also requires a **matching mode** before adopting, so a collaborator's same-name-but-different-mode branch in the skew window can't be adopted (test added); plus CodeRabbit's nitpick — a `withCapturedExit` test helper.

Net: the diff since your pass adds the mode guard + the timeout exit + test/lint tidy; nothing else changed. `create.test.ts` is 14/14, build + lint clean. Would appreciate your green-check on the final state when you have a moment — no rush.

jwfing (approved):

PR #206, opened 2026-07-23
feat: add feedback command for reporting InsForge-side hurdles

Fermionic-Lyu:
## Summary

Adds `insforge feedback` — agents (and humans) report InsForge-side hurdles mid-task: bugs, missing capabilities, or DX friction, anywhere in the toolkit.

```bash
npx @insforge/cli feedback --json \
  --type bug --component sdk --language python --area storage \
  --title "upload() hangs for files over 5MB" \
  --detail "<what happened vs expected, minimal repro>" \
  --error "<verbatim output>" --workaround "<what worked instead>"
```

### Taxonomy (designed backwards from triage)

| Flag | Values | Question |
|---|---|---|
| `--type` (required) | `bug` \| `feature-request` \| `friction` \| `other` | What kind of hurdle? |
| `--component` (required) | `backend` \| `sdk` \| `cli` \| `skills` \| `docs` \| `other` | Which deliverable? |
| `--area` (optional) | `db`, `auth`, `storage`, … | Which product feature? |

Docs-vs-behavior discrepancies file as `bug` with `--doc`/`--expected`/`--workaround` carrying the evidence. `--language` is required with `--component sdk`.

### Privacy & anti-spam

- **Local PII redaction** (`src/lib/redact.ts`): emails, JWTs, bearer tokens, known key formats, secret assignments, connection-string credentials, public IPs, home-dir usernames. Redact-then-truncate ordering; per-field caps.
- **No login required** — OSS and logged-out users can report. Abuse is contained server-side in a dedicated InsForge feedback project (dogfooding): RLS-locked table with a validating edge function as the only write path, 20/hour per-IP rate limit (salted hash, raw IP never stored), 7-day duplicate folding. The backend is maintained in the feedback project itself, not in this repo — see DEVELOPMENT.md §6 for how to inspect/redeploy it.
- **PostHog** gets metadata only (type/component/severity/lengths) — never free text, per DEVELOPMENT.md §2.

The endpoint URL and anon key are hardcoded public client credentials (rationale in `src/lib/api/feedback.ts`), overridable via `INSFORGE_FEEDBACK_URL` / `INSFORGE_FEEDBACK_ANON_KEY`.

## Test plan

- 23 new tests (`src/lib/redact.test.ts`, `src/commands/feedback.test.ts`); full suite 627 passing, lint clean.
- Verified end-to-end against the live backend: new report → 201 + id; duplicate → folds into existing row (`duplicate_count` bumped); anon key denied direct table read/write; real built CLI submission confirmed.

## Cross-references & merge order

Companion skills PR: [InsForge/insforge-skills#119](https://github.com/InsForge/insforge-skills/pull/119) routes agents here from the `insforge-cli`, `insforge`, and `insforge-debug` skills. **Publish this to npm first, then merge the skills PR** — skills install from the repo's default branch at runtime, so merging skills first would document a command agents don't have yet.

Follow-ups tracked outside this PR: rotate the feedback project admin key, assign a triage owner for `status = 'new'` rows.

> [!NOTE]
> ### Add `feedback` CLI command for reporting issues to InsForge
> - Adds a new `feedback` subcommand ([src/commands/feedback.ts](https://github.com/InsForge/CLI/pull/206/files#diff-48496a44b8103cdd7f5a9500af8dfcffe184e4f649a12c42ed403f12ca159ecd)) that collects structured reports (type, component, severity, area, etc.) and POSTs them to a hardcoded InsForge edge function endpoint via [`submitFeedback`](https://github.com/InsForge/CLI/pull/206/files#diff-2125049203abf10390a46767d8e4b2c75e6aac15f18fffc2baafb9ae7e50d7ff).
> - Supports interactive prompting for missing required fields in TTY mode, and JSON output mode for scripted use.
> - Free-text fields are sanitized via [`redactSensitive`](https://github.com/InsForge/CLI/pull/206/files#diff-cb4af16a85f20c9d2efc9d220be5a527429e2ce2e56c931c43f1b75a3fe4c0ca) (strips emails, JWTs, API keys, home dirs) and capped with `truncateMiddle` before submission.
> - Project context (project_id, org_id, region) is included in the payload unless running in OSS or unlinked mode. Analytics receive only enumerated values; free-text area/language maps to `'other'`.
> - Risk: the feedback endpoint and anon key are hardcoded with optional env overrides; rotating them requires a CLI release.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 6083b39.</sup>
> 
>

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -139,3 +139,24 @@ silently disabled in published artifacts.
 If a release goes wrong, deprecate the bad version with `npm deprecate` rather
 than unpublishing — unpublishing breaks downstream users who already installed
 that version.
+
+## 6. Feedback backend
+
+The `insforge feedback` command submits reports to a dedicated InsForge
+project via a public edge function — no login required, spam contained
+server-side (RLS-locked table, allow-list validation, per-IP rate limit,
+duplicate folding). The endpoint URL and anon key are deliberately
+hardcoded in `src/lib/api/feedback.ts` (public client credentials
+protected by RLS + function-side rate limiting), overridable via
+`INSFORGE_FEEDBACK_URL` / `INSFORGE_FEEDBACK_ANON_KEY` for testing and
+rotation.
+
+The backend itself (the `feedback` table, the `submit-feedback` edge
+function, and its secrets) lives in the feedback project — it is not
+versioned in this repo. To inspect or maintain it, link a scratch
+directory to that project with the admin key, then use
+`insforge functions code submit-feedback`, `db query`, and
+`functions deploy` as usual; the project's agent memory
+(`insforge memory list`) records the design decisions and triage
+workflow. If you change the CLI's feedback payload shape, update the
+function's allow-list and the table schema in the same change set.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@insforge/cli",
-  "version": "0.2.0",
+  "version": "0.2.1",
   "description": "InsForge CLI - Command line tool for InsForge platform",
   "type": "module",
   "bin": {
--- a/src/commands/feedback.test.ts
+++ b/src/commands/feedback.test.ts
@@ -1,0 +1,270 @@
+import { describe, it, expect, vi, beforeEach, type Mock } from 'vitest';
+import { Command } from 'commander';
+import { registerFeedbackCommand } from './feedback.js';
+import { resetTelemetryOutcomeForTests } from '../lib/command-telemetry.js';
+
+vi.mock('../lib/api/feedback.js', () => ({
+  submitFeedback: vi.fn(async () => ({ id: 'fb_123', status: 'received' })),
+}));
+vi.mock('../lib/config.js', () => ({
+  FAKE_PROJECT_ID: 'fa4e0000-1234-5678-90ab-cd1234567890',
+  getProjectConfig: vi.fn(() => ({
+    project_id: 'p1',
+    project_name: 'demo',
+    org_id: 'o1',
+    appkey: 'k',
+    region: 'us-east',
+    api_key: 'key',
+    oss_host: 'http://localhost',
+  })),
+}));
+vi.mock('../lib/analytics.js', () => ({
+  trackGroupCommand: vi.fn(),
+  trackTopLevelCommand: vi.fn(),
+  shutdownAnalytics: vi.fn(async () => {}),
+}));
+
+function makeProgram() {
+  const program = new Command().exitOverride();
+  program.option('--json').option('--api-url <url>');
+  registerFeedbackCommand(program);
+  return program;
+}
+
+async function run(argv: string[]) {
+  const logSpy = vi.spyOn(console, 'log').mockImplementation(() => {});
+  try {
+    await makeProgram().parseAsync(argv, { from: 'user' });
+  } finally {
+    logSpy.mockRestore();
+  }
+}
+
+describe('feedback command', () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    resetTelemetryOutcomeForTests();
+  });
+
+  it('submits a structured payload with auto-attached context', async () => {
+    const { submitFeedback } = await import('../lib/api/feedback.js');
+
+    await run([
+      'feedback',
+      '--type', 'bug',
+      '--component', 'backend',
+      '--title', 'db policies create 500s on uppercase table names',
+      '--detail', 'Creating an RLS policy on table "Users" returns 500. Lowercase names work.',
+      '--area', 'db',
+      '--command', 'insforge db policies create ...',
+      '--severity', 'major',
+    ]);
+
+    const [payload] = (submitFeedback as Mock).mock.calls[0];
+    expect(payload).toMatchObject({
+      type: 'bug',
+      component: 'backend',
+      severity: 'major',
+      area: 'db',
+      project_id: 'p1',
+      org_id: 'o1',
+      region: 'us-east',
+    });
+    expect(payload.client_info).toMatchObject({
+      source: 'cli',
+      node_version: process.version,
+    });
+    expect(payload.client_info.os).toContain(process.platform);
+  });
+
+  it('redacts PII from free-text fields before sending', async () => {
+    const { submitFeedback } = await import('../lib/api/feedback.js');
+
+    await run([
+      'feedback',
+      '--type', 'bug',
+      '--component', 'backend',
+      '--title', 'auth fails for jane@example.com',
+      '--detail', 'Login with key uak_a1b2c3d4e5f6 under /Users/jane/app fails.',
+      '--error', 'Error: invalid token eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxIn0.abc123def456',
+    ]);
+
+    const [payload] = (submitFeedback as Mock).mock.calls[0];
+    expect(payload.title).toBe('auth fails for [REDACTED_EMAIL]');
+    expect(payload.detail).toBe('Login with key [REDACTED_KEY] under ~/app fails.');
+    expect(payload.error).toContain('[REDACTED_JWT]');
+    expect(JSON.stringify(payload)).not.toContain('jane');
+  });
+
+  it('rejects a missing type in non-interactive mode with a usage hint', async () => {
+    const errSpy = vi.spyOn(console, 'error').mockImplementation(() => {});
+    const exitSpy = vi.spyOn(process, 'exit').mockImplementation((() => {
+      throw new Error('exit');
+    }) as never);
+    try {
+      await expect(
+        run(['feedback', '--component', 'cli', '--title', 't', '--detail', 'd']),
+      ).rejects.toThrow('exit');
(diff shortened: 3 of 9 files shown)


jwfing (changes requested):
## Review: `feat: add feedback command`

**Summary:** A well-structured, well-tested `insforge feedback` command with a thoughtful local-redaction + server-side anti-spam design; the one blocking issue is that the committed edge-function source passes an option to `createClient` that the InsForge SDK does not appear to recognize, which would break every real submission.

### Requirements context

No matching spec/plan found under a `docs/superpowers/` (absent) or `docs/specs/` directory — the specs present (`diagnose`, `db-migrations`) don't cover this feature. Assessed against the PR description, `DEVELOPMENT.md`, and the `feedback-backend/README.md` design notes. The SDK query-builder / `createClient` API was cross-checked against the live InsForge docs via context7.

---

### 🔴 Critical

**Functionality — edge function likely never authenticates as admin** — `feedback-backend/functions/submit-feedback.ts:102-105`

```ts
const client = createClient({
  baseUrl: Deno.env.get('INSFORGE_BASE_URL'),
  accessToken: Deno.env.get('FEEDBACK_SERVICE_KEY'),
});
```

`accessToken` is not a documented `createClient` option. The InsForge SDK docs consistently show `anonKey` for public/unauthenticated clients and `edgeFunctionToken` for token-based server-side access inside edge functions (and the service-role pattern passes the privileged key via the key field, not `accessToken`). If the SDK ignores unknown config keys, this client is created **without any credential**, so its reads/writes against `public.feedback` — which is RLS-locked with zero policies and `revoke all ... from anon, authenticated` (`migrations/20260723201839_create-feedback.sql:54-55`) — will be denied. The rate-limit `select` then errors → `500 'Storage unavailable'`, or the `insert` fails → `500 'Could not store feedback'`, on **every** real submission. Only the `OPTIONS`/validation/`413` paths would work.

This directly contradicts the PR's "verified end-to-end … new report → 201 + id" claim, which strongly suggests the **deployed** function differs from the **committed** source. That is itself a problem: `feedback-backend/README.md` tells a future maintainer to redeploy from this file (`functions deploy submit-feedback --file functions/submit-feedback.ts`), so shipping a source that wasn't the thing verified means the next redeploy pushes broken code.

Please reconcile: confirm the exact `createClient` field the SDK uses for a service/admin key server-side (likely `edgeFunctionToken` or the key field, not `accessToken`), and make the committed source byte-for-byte the version that returns 201 against the live backend.

---

### 🟡 Suggestion

- **Test coverage — the backend logic is untested** (`feedback-backend/functions/submit-feedback.ts`). The CLI side is well covered (23 tests across `redact.test.ts` / `feedback.test.ts`), but the server-side rate limit, duplicate folding, IP hashing, field allow-listing, and required-field checks have no tests. The Critical above is exactly the class of bug a single "submits and returns 201" integration test would have caught. Even a lightweight harness (or a documented manual test script checked into `feedback-backend/`) would raise confidence, given this is the only write path.

- **Security — `x-forwarded-for` uses the client-controllable leftmost hop** (`submit-feedback.ts:93`). `.split(',')[0]` takes the first XFF entry, which a caller can forge by sending their own `X-Forwarded-For` header — bypassing both the per-IP rate limit and the salted-hash IP privacy (a spoofed value just partitions the counter). Unless the InsForge platform proxy strips/overwrites inbound XFF before the function sees it, prefer the right-most trusted hop (or a platform-provided real-IP header). Worth confirming what the platform guarantees here.

- **Functionality/perf — rate limit and dedupe are read-then-write, non-atomic** (`submit-feedback.ts:110-137`). Two concurrent submissions from the same IP can both pass the `>= RATE_LIMIT_PER_HOUR` check, and two concurrent duplicates can both miss the dedupe `select` and insert two rows. Blast radius is low for a feedback endpoint, but an agent retry-loop firing in parallel is precisely the abuse case the design calls out. A unique index on `dedupe_hash` (within window) or an upsert/RPC would make folding race-free.

### 🔵 Information

- **`MAX_BODY_BYTES` compares character count, not bytes** (`submit-feedback.ts:67`): `raw.length` is UTF-16 code units, so a multibyte (non-ASCII) payload can exceed 32 KB of actual bytes before this trips. Minor; the field caps still bound stored size.

- **Migration `revoke` is unguarded** (`migrations/...:55`): `revoke all ... from anon, authenticated` assumes both roles exist. They always do in a standard InsForge project (PostgREST roles), so this is safe here — noting only because per repo convention role grants/revokes are sometimes guarded with `IF EXISTS (select from pg_roles ...)` to keep a migration chain from aborting if ever run against a project missing them.

- **Backend does not validate `type`/`component` enum values** (by design, per the migration comment — DB only bounds length so a newer CLI can add values without a backend migration). Consequence: a direct POST bypassing the CLI can store arbitrary classification strings. Acceptable given the documented rationale; flagging for awareness since dedupe keys off `type|component|title`.

- Convention checks pass: command follows the `registerXxxCommand` pattern and top-level registration in `src/index.ts`; `fetch` lives in an `api/` wrapper (`src/lib/api/feedback.ts`) consistent with `apify.ts`/`posthog.ts`; errors go through `CLIError`/`handleError`; `--json` suppresses prompts; analytics uses the shared `trackTopLevelUsage` helper and sends metadata only (verified by `feedback.test.ts` "sends only metadata to analytics"); `CLI_VERSION` is build-injected via `tsup.config.ts`. Local redaction (redact-then-truncate ordering) and its test suite are solid.

---

One Critical item (the `createClient` credential option) blocks merge — please confirm the correct SDK field and ensure the committed backend source matches what was verified. Everything else is non-blocking. Nice work on the redaction layer and the layered anti-spam design.

jwfing (commented):
## Review: `feat: add feedback command`

**Summary:** A clean, well-tested addition of `insforge feedback` (command + local PII redaction + edge-function client) that faithfully implements the PR description; no blocking issues found.

### Requirements context
No matching spec/plan found — `docs/specs/` contains only `diagnose` and `db-migrations` designs, and there is no `docs/superpowers/` in this repo. Assessed against the PR description, `DEVELOPMENT.md §2` (analytics rules) and the newly added `§6` (feedback backend), and surrounding CLI conventions.

---

### Critical
(none)

---

### Suggestions

**Security / redaction coverage — `src/lib/redact.ts:12-30`**
The known-key allow-list catches InsForge `uak_`, Stripe, OpenAI `sk-`, GitHub, AWS `AKIA`, and Slack `xox*`, but bare high-entropy tokens in other common formats slip through unless they appear in `key=value` form: e.g. Google `AIzaSy…`, `npm_…`, PostHog `phc_…`. I verified this by running the regex table against those fixtures — they pass through unredacted. Blast radius is genuinely low (the destination is InsForge's own re-scrubbing backend, PostHog receives metadata only, and the module is explicitly documented as a conservative *first* line of defense), so this is a suggestion rather than a blocker — but since local scrubbing is the whole point of the file, adding a couple more prefixes is cheap insurance.

**Test coverage — `src/commands/feedback.test.ts`**
Strong suite (23 tests, edge cases well chosen). Three uncovered branches would round it out:
- `--file <path>` detail-from-file reading (`feedback.ts:169`) has no test.
- OSS detection is tested via `FAKE_PROJECT_ID` but not the *unlinked* path where `getProjectConfig()` returns `null` (`feedback.ts:212` — `!config` branch).
- The `status === 'duplicate'` user-facing output branch (`feedback.ts:257`) isn't exercised (only the API-layer mapping is, in `feedback.api.test.ts`).

---

### Information

- **`src/lib/redact.ts:22-25`** — the generic secret-assignment pattern requires a value of 6+ chars, so `password="ab"`-style short secrets pass through. Acceptable given the conservative-by-design intent; noting only.
- **`DEVELOPMENT.md §6`** — clearly documents the out-of-repo backend and the "keep the function allow-list in sync with the payload shape" invariant. Good.

---

### Dimension notes
- **Software engineering** — TDD followed; `.js` ESM extensions and `node:` import prefix match repo convention (`storage/upload.ts`); imports all used; telemetry uses the shared `trackTopLevelUsage` helper per `§2`, not the retired `reportCliUsage`.
- **Functionality** — Implements the described taxonomy, `--language` requirement for SDK, redact-then-truncate ordering, and OSS-mode context omission. `getProjectConfig()` returns `null` (doesn't throw) for unlinked users, so no-login submission works as claimed. `CLI_VERSION` is correctly build-injected via `tsup.config.ts` (same pattern as `diagnose`).
- **Security** — Hardcoded endpoint + anon key are public client credentials by design (RLS-locked table, function-only write path), documented and env-overridable; no real secret is committed. Analytics props are all enums/booleans/counts — free text (`title`/`detail`/`area`/`language`) is either excluded or normalized to `other` via `telemetryEnum`, satisfying `§2`; `sanitizeTelemetry` truncation is a further backstop. Timeout rejection is a `DOMException` that *is* `instanceof Error` with `name === 'TimeoutError'` on Node ≥18 (verified), so the friendly timeout message path fires in production.
- **Performance** — Single `fetch` with a 10s `AbortSignal.timeout`; no loops, N+1, or blocking work. No concerns.

---

### Verdict: **approved** (informational — a human still gives the explicit GitHub approval)
No Critical findings. The Suggestion and Information items are non-blocking; merge order note (npm publish before the skills PR) is sound.

jwfing (dismissed):
LGTM - approved.

jwfing (approved):
LGTM - approved.

PR #207, opened 2026-07-26
feat: telemetry opt-out command; remove hidden records commands

Fermionic-Lyu:
## Summary

Two changes:

### 1. `insforge telemetry` — usage-analytics opt-out

```bash
npx @insforge/cli telemetry status    # enabled/disabled + what decides it
npx @insforge/cli telemetry disable   # persists to ~/.insforge/config.json
npx @insforge/cli telemetry enable
```

Design:

- **One kill switch, every emitter.** `isTelemetryDisabled()` in `src/lib/analytics.ts` gates both the PostHog client and the legacy `reportCliUsage` OSS path. DEVELOPMENT.md §2 now requires any new telemetry emitter to check it.
- **Resolution order:** `DO_NOT_TRACK` (the [cross-tool convention](https://consoledonottrack.com)) → `INSFORGE_TELEMETRY_DISABLED` (per-run/CI) → `telemetry_disabled` in the global config (persistent, managed by this command). Explicit falsy values (`0`, `false`, empty) don't opt out.
- **Honest status:** `telemetry enable` errors if an env var still forces telemetry off, instead of claiming it's back on; `status` reports the deciding source.
- **No self-tracking:** the telemetry command emits no analytics events — an opt-out flow that phones home undermines the point.
- **Fail-safe:** a corrupt config file reads as "enabled"; telemetry handling can never break a command.

`reportAgentConnected` (dashboard "agent connected" onboarding signal) is deliberately not gated — it drives product behavior, not analytics.

### 2. Remove the hidden `records` command group

Never supported for direct use (hidden since introduction); table data goes through `db query`. Deletes `src/commands/records/`, the registration in `src/index.ts`, and updates the README note.

## Test plan

- 12 new tests: env-var precedence (including falsy values), config opt-out, corrupt-config fail-safe, PostHog gating, and the command's status/enable/disable behaviors including the env-override error path.
- Full suite: 648 passing, lint clean, build clean.
- Smoke-tested on the built CLI: `records` errors as an unknown command; `telemetry status --json` reports `default`/`INSFORGE_TELEMETRY_DISABLED`/`config` sources correctly; disable→enable round-trip verified against a real `~/.insforge/config.json`.

---
## Summary by cubic
Adds `insforge telemetry` to manage anonymous usage analytics with a single opt‑out switch. Removes the hidden `records` commands; use `db query` for table data.

- **New Features**
  - One kill switch for telemetry. Resolution: `DO_NOT_TRACK` → `INSFORGE_TELEMETRY_DISABLED` → `telemetry_disabled` in `~/.insforge/config.json`.
  - Gates both PostHog and the legacy `reportCliUsage`. New emitters must check it (see DEVELOPMENT.md).
  - `telemetry status|enable|disable`; `status` explains the deciding source, and `enable` warns if an env var still keeps it off. The command does not self-track.
  - Corrupt config fails safe (telemetry defaults to enabled so commands never break).

- **Refactors**
  - Remove the hidden `records` group (list/create/update/delete) and its registration. Docs now direct table access through `db query`.

<sup>Written for commit 8b991d6dffecffbd7ed3a4cb32b9ec5bb48e6493. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/CLI/pull/207?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Add `telemetry` CLI command with opt-out support and remove hidden `records` commands
> - Adds a new `telemetry` command group with `status`, `enable`, and `disable` subcommands that persist opt-out state in the global config via a `telemetry_disabled` flag in `GlobalConfig`.
> - Introduces `isTelemetryDisabled()` in [analytics.ts](https://github.com/InsForge/CLI/pull/207/files#diff-af25af6c1c85456cbce5a4eda14f0b3965aacdd964420dde9a08ee5c7c2674af) as a centralized kill switch that checks `DO_NOT_TRACK` and `INSFORGE_TELEMETRY_DISABLED` env vars and the persisted config; both `getClient()` and `reportCliUsage()` now respect it.
> - Removes the hidden `records` CRUD command group (`create`, `delete`, `list`, `update`) from the CLI; users are directed to `db query` instead.
> - Behavioral Change: PostHog events and legacy usage reporting are silently suppressed when telemetry is disabled by env or config.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 8b991d6.</sup>
> 
>

## Summary by CodeRabbit

* **New Features**
  * Added `telemetry status`, `telemetry enable`, and `telemetry disable` CLI commands.
  * Added persistent and per-run controls for anonymous usage analytics, including environment-variable overrides.
* **Changes**
  * Removed the `records` CLI commands for listing, creating, updating, and deleting records. Use `db query` instead.
* **Documentation**
  * Updated command references, environment variable documentation, and telemetry configuration guidance.
* **Tests**
  * Added coverage for telemetry status, opt-out behavior, configuration persistence, and environment overrides.


jwfing (commented):
## Review: telemetry opt-out + records removal

**Summary:** A clean, well-tested telemetry opt-out implemented as a single kill switch that correctly gates every analytics emitter, plus a tidy removal of the dead hidden `records` group — no blocking issues.

### Requirements context
No matching spec/plan found under `docs/` — the repo keeps specs in `docs/specs/` (only `diagnose` and `db-migrations` design docs exist; there is no `docs/superpowers/`). Assessed against the PR description, `DEVELOPMENT.md §2`, and the surrounding code.

### Findings

#### Critical
(none)

#### Suggestion

- **Software engineering — duplicated falsy-env parsing (`src/commands/telemetry.ts:9-15` vs `src/lib/analytics.ts:8-31`).** `telemetry.ts` reimplements the "set to anything except an explicit falsy value" rule (`envOverride()`/`flag()`) that already lives in `analytics.ts` (`envFlag()`/`isTelemetryDisabled()`). The command needs the *source name* (`DO_NOT_TRACK` vs `INSFORGE_TELEMETRY_DISABLED`), which `isTelemetryDisabled()` doesn't expose, so the duplication is understandable — but two independent copies of the same precedence + falsy rule can drift, and if they do, `telemetry status` would report a `source` that disagrees with what actually gates emission. Consider exporting a single resolver from `analytics.ts` (e.g. one that returns `{disabled, source}`) and having both `isTelemetryDisabled()` and the command consume it.

- **Functionality — corrupt config isn't fail-safe in the command path (`src/commands/telemetry.ts:33-38`, `38-50`, `52-73`).** `isTelemetryDisabled()` deliberately treats a corrupt `~/.insforge/config.json` as "enabled" (try/catch), so telemetry keeps flowing. But `status`/`disable`/`enable` call `getGlobalConfig()` unguarded, and `config.ts:29-30` `JSON.parse`s with no try/catch — so with a corrupt config, `telemetry disable` throws and exits instead of opting the user out. Net effect: the one state where telemetry is silently on (corrupt config → fail-safe "enabled") is also the state where the persistent opt-out command fails, leaving only the env var as an escape. Low blast radius, but mirroring the PR's own fail-safe stance (e.g. `disable` starting from a default config when the file is unreadable) would be more robust.

#### Information

- **Performance — `isTelemetryDisabled()` re-reads config per emission (`src/lib/analytics.ts:23-31`, `36`).** `getClient()` now calls `getGlobalConfig()` → synchronous `readFileSync` + `JSON.parse` on every telemetry event, and the result isn't memoized even though the PostHog `client` is cached. Negligible for a one-shot CLI (typically one event per invocation), noting only in case an event ever fires in a loop.

- **Security — no concerns.** No new user input reaches SQL/shell/HTTP; the command only writes a boolean to config. No secrets/PII newly logged. This PR is net-positive for privacy (adds opt-out) and the `records` removal shrinks the OSS-fetch surface. No new dependencies.

- **Software engineering — coverage & conventions are solid.** 12 new tests exercise env precedence (including explicit-falsy `0`/`false`), config opt-out, corrupt-config fail-safe, PostHog gating, and the `enable` env-override error path. Import style (`.js` suffixes, `node:` prefixes), error handling (`handleError`/`CLIError`), and the "no self-tracking" note are consistent with the codebase. The kill switch is genuinely centralized — verified `command-telemetry.ts` (`trackCommandUsage`/`trackTopLevelUsage`) routes through `captureEvent → getClient()`, so it's gated too, and no orphaned `records` references remain after the deletion. `DEVELOPMENT.md` now documents the opt-out contract for future emitters.

### Verdict
**approved** (no Critical findings). Suggestions above are non-blocking. Note: this is the bot's assessment for reporting — the GitHub green-check approval remains a separate human action.

jwfing (approved):
LGTM - approved.

PR #215, opened 2026-07-30
feat: add --toon output format for token-optimized CLI output

akb4q:
## Summary

Add a global `--toon` flag that outputs results in **TOON** (Token-Oriented Object Notation), designed for LLM consumption — 6-9x more compact than JSON or human-readable tables.

## Performance

Measured on `insforge projects list`:
| Format | Size | vs JSON |
|--------|------|---------|
| TOON | 129 B | **7.7× smaller** |
| JSON | 989 B | 1× (baseline) |
| Human table | 1,162 B | 0.85× |

## Usage

```
insforge projects list --toon
# [2]{ID,Name,Region,Status,AppKey}:
#   uuid-1,my-project,us-east,active,key-1
#   uuid-2,other-project,eu-west,active,key-2

insforge whoami --json --toon
# id: 3db0ba25-...
# name: Alice
# email: alice@example.com

insforge functions list --toon
# [3]{Slug,Name,Status,"Created At"}:
#   func-1,My Function,active,"2026-07-30, 10:00:00 AM"
```

## Implementation

- **TOON format**: toonformat.dev — token-optimized notation with 36+ language implementations
- **Module-level flag** set via Commander `preAction` hook, read by `outputTable` and `outputJson` at render time — zero changes to any command handler
- **Spec compliance**: `toonEscapeValue` / `toonEscapeKey` conform to TOON SPEC §7.1-7.3 (escaping, quoting rules, key encoding)
- **Non-structured suppression**: `outputSuccess`/ `outputInfo` silently skipped in TOON mode to keep output parseable

## Testing

- `npm run build`: tsup compiles clean (601 KB ESM)
- `npm test`: 649 tests pass (1 pre-existing Cloudflare OAuth failure unrelated)
- Manual integration: `projects list`, `whoami`, `functions list`, `orgs list` with `--toon` and `--json --toon`

Root commands (create, login, feedback, etc.) produce no TOON output — they emit interactive prompts and status messages, not structured data.

---
## Summary by cubic
Adds a global `--toon` output format for token‑optimized results, ~6–9× smaller than JSON or tables. Uses `@toon-format/toon` `encode()` and suppresses non‑structured messages to keep output parseable.

- **New Features**
  - Global `--toon` via pre‑action hook; `outputJson` uses `encode()` and tables render as tabular TOON.
  - Spec‑compliant escaping for headers/values; interactive root commands unchanged.

- **Bug Fixes**
  - Treat `--toon` as implying JSON in `getRootOpts` to avoid empty output and credential loss.
  - Quote structural characters correctly and suppress `outputSuccess`/`outputInfo` in TOON mode; add tests for quoting, nulls, single‑doc output, and empty tables.

<sup>Written for commit 0b66a55626b4b467c3befaa999931a112c86b70a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/CLI/pull/215?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Add `--toon` global CLI flag for token-optimized TOON format output
> - Adds a `--toon` global flag to the CLI that switches all output to [TOON format](https://toonformat.dev) using the `@toon-format/toon` package.
> - In TOON mode, `outputJson` encodes data via `encode()`, `outputTable` renders a tabular TOON document with header schema and row values, and `outputSuccess`/`outputInfo` are silently suppressed.
> - Updates `getRootOpts().json` in [errors.ts](https://github.com/InsForge/CLI/pull/215/files#diff-33a1fa952d3520ae3b8ff121fe89f8d401e10efd4cbf34f180663caa16c0f153) to return `true` when `--toon` is active, so error-handling code treats TOON mode like JSON mode.
> - Behavioral Change: success and info messages are fully suppressed when `--toon` is used, which may hide feedback in interactive workflows.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 0b66a55.</sup>
> 
>

## Summary by CodeRabbit

- **New Features**
  - Added a global `--toon` option for structured TOON-formatted command output.
  - JSON and table responses now support TOON formatting.
  - TOON mode is applied consistently across command actions and error handling.
- **Bug Fixes**
  - Improved consistency between TOON and standard JSON output behavior.
  - Success and informational messages are suppressed in TOON mode to preserve structured output.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
       "dependencies": {
         "@clack/prompts": "^0.9.1",
         "@insforge/shared-schemas": "^1.1.58",
+        "@toon-format/toon": "^4.1.0",
         "archiver": "^7.0.1",
         "cli-table3": "^0.6.5",
         "commander": "^13.1.0",
@@ -1475,6 +1476,12 @@
       "resolved": "https://registry.npmjs.org/@standard-schema/spec/-/spec-1.1.0.tgz",
       "integrity": "sha512-l2aFy5jALhniG5HgqrD6jXLi/rUWrKvqN/qJx6yoJsgKhblVd+iqqU4RCXavm/jPityDo5TCvKMnpjKnOriy0w==",
       "dev": true,
+      "license": "MIT"
+    },
+    "node_modules/@toon-format/toon": {
+      "version": "4.1.0",
+      "resolved": "https://registry.npmjs.org/@toon-format/toon/-/toon-4.1.0.tgz",
+      "integrity": "sha512-dBB3pkEx9QYvHnHR6rtkaBAh+7x4W/oA5ONur4G0fh7Ow69PbPuM7OFxzNRABqyxC0t6SZ3RixiGbCuaFjPDAQ==",
       "license": "MIT"
     },
     "node_modules/@tybys/wasm-util": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@insforge/cli",
-  "version": "0.2.0",
+  "version": "0.2.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@insforge/cli",
-      "version": "0.2.0",
+      "version": "0.2.1",
       "license": "Apache-2.0",
       "dependencies": {
         "@clack/prompts": "^0.9.1",
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
   "dependencies": {
     "@clack/prompts": "^0.9.1",
     "@insforge/shared-schemas": "^1.1.58",
+    "@toon-format/toon": "^4.1.0",
     "archiver": "^7.0.1",
     "cli-table3": "^0.6.5",
     "commander": "^13.1.0",
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,7 +5,7 @@ import { Command } from 'commander';
 import * as clack from '@clack/prompts';
 import * as prompts from './lib/prompts.js';
 import { getCredentials, getProjectConfig } from './lib/config.js';
-import { outputJson } from './lib/output.js';
+import { outputJson, setToonMode } from './lib/output.js';
 import { registerLoginCommand } from './commands/login.js';
 import { registerLogoutCommand } from './commands/logout.js';
 import { registerWhoamiCommand } from './commands/whoami.js';
@@ -116,6 +116,7 @@ program
 // Global options
 program
   .option('--json', 'Output in JSON format')
+  .option('--toon', 'Output in TOON format (token-optimized notation, see toonformat.dev)')
   .option('--forger', 'Play the Forger animation (root command only) and return to the interactive menu')
   .option('--api-url <url>', 'Override Platform API URL')
   .option('-y, --yes', 'Skip confirmation prompts')
@@ -152,6 +153,12 @@ program.hook('preAction', async (_thisCommand, actionCommand) => {
 // operations for human approval. Lives in the CLI so it protects every caller
 // (Claude Code, Cursor, scripts, CI, humans) automatically.
 program.hook('preAction', guardHook);
+
+// TOON output format: set module-level flag before any action runs
+program.hook('preAction', (thisCommand: Command, actionCommand: Command) => {
+  const opts = actionCommand.optsWithGlobals() as { toon?: boolean };
+  setToonMode(!!opts.toon);
+});
 
 // Top-level commands
 registerLoginCommand(program);
--- a/src/lib/errors.ts
+++ b/src/lib/errors.ts
@@ -126,7 +126,7 @@ export function getRootOpts(cmd: Command): { json: boolean; apiUrl?: string; yes
   }
   const opts = root.opts();
   return {
-    json: opts.json ?? false,
+    json: opts.json ?? opts.toon ?? false,
     apiUrl: opts.apiUrl,
     yes: opts.yes ?? false,
   };
--- a/src/lib/output.test.ts
+++ b/src/lib/output.test.ts
@@ -1,0 +1,214 @@
+import { describe, it, expect } from 'vitest';
+import { setToonMode, outputTable, outputSuccess, outputInfo, outputJson } from './output.js';
+
+describe('toonEscapeValue (indirectly via outputTable)', () => {
+  it('quotes values containing structural chars (comma, colon, bracket, etc.)', () => {
+    // outputTable with toon mode writes TOON to stdout
+    // We capture console.log to test the output
+    const logs: string[] = [];
+    const origLog = console.log;
+    console.log = (msg: string) => logs.push(msg);
+
+    try {
+      setToonMode(true);
+      outputTable(['Name'], [['Smith, John']]);
+      expect(logs[0]).toContain('"Smith, John"');
+    } finally {
+      console.log = origLog;
+      setToonMode(false);
+    }
+  });
+
+  it('outputTable renders colon-containing values quoted', () => {
+    const logs: string[] = [];
+    const origLog = console.log;
+    console.log = (msg: string) => logs.push(msg);
+
+    try {
+      setToonMode(true);
+      outputTable(['Key'], [['a:b']]);
+      expect(logs[0]).toContain('"a:b"');
+    } finally {
+      console.log = origLog;
+      setToonMode(false);
+    }
+  });
+
+  it('does not quote simple alphanumeric values', () => {
+    const logs: string[] = [];
+    const origLog = console.log;
+    console.log = (msg: string) => logs.push(msg);
+
+    try {
+      setToonMode(true);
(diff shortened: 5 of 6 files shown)


akb4q:
Closes #216

jwfing (changes requested):
## Review: `feat: add --toon output format`

**Summary:** Clean, well-scoped feature (module-level flag routed through `outputTable`/`outputJson`, zero handler changes), but a broken regex in the value-quoting logic means values containing the delimiter or other structural characters are emitted **unquoted**, producing malformed TOON — including on the PR's own headline example.

### Requirements context
No matching spec/plan found — this repo keeps specs under `docs/specs/` (not `docs/superpowers/`), and none of the three present (`diagnose`, `db-migrations`) cover output formatting. Assessed against the PR description and TOON's stated goal (compact, safely-parseable output for LLM/script consumption).

---

### Critical

**1. Value-quoting regex closes its character class early → delimiter and structural chars are never quoted (`src/lib/output.ts:23`)**

```js
/[:,"\\{}[\\]]/.test(s)
```

The intent is "quote if the value contains any of `: , " \ { } [ ]`". But inside the class, `\\` is a literal backslash and the very next `]` **closes the class**. So the pattern actually parses as `[:,"\{}[\]` followed by a trailing literal `]` — i.e. it only matches when one of those chars is *immediately followed by* a `]`. Verified:

```
"a,b"        -> match: false   // comma NOT detected
"a:b"        -> match: false
'a"b'        -> match: false
"a{b}c"      -> match: false
"Smith, John"-> match: false
":]"         -> match: true    // only the accidental trailing-] form matches
```

Consequence: any value containing a **comma** (the TOON row delimiter) is emitted unquoted, so the row gains phantom columns and the output no longer parses. This is not hypothetical — it hits the PR's own advertised example. `functions list` builds the `Created At` cell via `new Date(f.createdAt).toLocaleString()` (`src/commands/functions/list.ts:45`), which in en-US yields e.g. `7/30/2026, 10:00:00 AM` (contains a comma). Actual output:

```
[3]{Slug,Name,Status,"Created At"}:
  func-1,My Function,active,7/30/2026, 10:00:00 AM   <- 5 fields for a 4-col header
```

whereas the PR body shows the timestamp correctly quoted (`"2026-07-30, 10:00:00 AM"`) — output the current code does **not** produce. Since the whole point of `--toon` is safely machine-parseable output, this defeats the feature and silently corrupts data.

Fix: escape the closing bracket so it stays inside the class, e.g. `/[:,"\\{}[\]]/` (verified: this correctly quotes `Smith, John`). Please also cover `"`, `\`, `{`, `}`, `[`, `]` while you're at it — all currently slip through the same way.

---

### Suggestion

- **Software engineering — no tests added (`src/lib/output.ts`).** This module is regex- and edge-case-heavy, and the repo has a strong test culture (600+ tests, one `*.test.ts` per lib module). A one-line assertion like `expect(toonEscapeValue('a,b')).toMatch(/^"/)` would have caught the Critical above immediately. Please add a `src/lib/output.test.ts` covering: delimiter/structural-char quoting, empty string, numeric-looking strings, booleans/null, control chars, leading/trailing whitespace, `[N]{keys}:` tabular framing, the `[0]{keys}:` empty case, single-object key/value, and the `--json --toon` path. The `insforge-development` skill requires every new behavior to carry at least one test.
- **Functionality — heterogeneous arrays (`src/lib/output.ts:102`).** `outputJsonToon` derives columns from `Object.keys(first)` only; if later array elements have different keys, extra keys are dropped and missing ones render as `null`. Fine for uniform list endpoints, but worth a guard or a comment documenting the assumption.

### Information

- **`--json --toon` precedence (`src/lib/output.ts:50`).** When both flags are passed, TOON silently wins and `--json` is ignored. The PR body uses `whoami --json --toon` as an intended combo, so this is presumably deliberate — worth a note in `--help` so callers aren't surprised.
- **Nested objects (`src/lib/output.ts:120-123`).** In single-object mode, nested objects are rendered as raw `JSON.stringify(v)` (embedded JSON, not TOON). Acceptable, but not strictly spec-conformant if downstream expects pure TOON.
- **`toonEscapeKey` (`src/lib/output.ts:37-45`)** doesn't apply the control-char `\uXXXX` catch-all that `toonEscapeValue` does. Keys come from code-controlled headers so the risk is negligible today.

### Security
No security-relevant changes — output-only path, no new user input reaching SQL/shell/HTTP, no new dependencies (formatting is hand-rolled), and TOON suppresses `outputSuccess`/`outputInfo` without touching secret handling.

### Performance
No concerns — string concatenation over CLI-sized result lists, no DB queries, no hot path or event-loop blocking introduced.

---

One Critical correctness bug (the quoting regex) blocks merge; it corrupts the exact output the feature exists to produce and lands untested. The rest is a solid, well-contained change — fixing the regex and adding a focused `output.test.ts` should be enough to land it.

jwfing:
@akb4q could you please address these comments? thanks!

akb4q:
@jwfing all review comments addressed:

- ✨ **Official encode**: Replaced ad-hoc serializer with @toon-format/toon `encode()` API — handles nested structures, heterogeneous arrays, and null values correctly
- 🔧 **`--toon` standalone**: `--toon` now works without `--json` — `getRootOpts()` routes to structured output automatically (commands like `whoami --toon`, `secrets rotate --toon` produce proper TOON output)
- 🔐 **One-time credentials**: `outputSuccess`/`outputInfo` emit structured TOON instead of suppressing — no more silent credential loss on `secrets rotate` or `s3-keys create`

Build ✅, ESLint 0 errors ✅, 648/649 tests passing ✅ (1 pre-existing Cloudflare flake)

akb4q:
@jwfing addressed your review and the new greptile re-review:

- **outputSuccess/outputInfo** now suppress in TOON mode. The `getRootOpts()` route (`--toon` implies `json: true`) ensures credential commands like `secrets rotate --toon` go through `outputJson(result)` which produces exactly one TOON document. No more multi-document stream.
- **Quoting regex** fixed. The original `/[:,"\\{}[\]]/` had `]` closing the character class early, so commas and colons were never detected as quote triggers. Fixed to `/[\]:,"\\{}[]/` using proper `\]` escape. Verified: `Smith, John`, `7/30/2026, 10:00:00 AM` are now correctly quoted.
- **Added `src/lib/output.test.ts`** with 13 tests: structural-char quoting, empty string, boolean-likes, null preservation, single-doc output, suppress behavior, tabular format, empty-table edge case.

Build passes, ESLint 0 errors, 661/662 tests passing (1 pre-existing Cloudflare flake).

jwfing (changes requested):
**Summary**
This PR is close, but the current implementation does not make `--toon` a truly global structured-output mode.

**Requirements context**
I did not find a linked issue or existing repo docs for TOON itself, so I assessed intent from the PR description plus existing repo conventions. The description says `--toon` is a global output flag with "zero changes to any command handler," and [DEVELOPMENT.md](/tmp/codex-review-GcGMbL/CLI/DEVELOPMENT.md:21) says structured output should flow through `outputJson` / `outputTable` / `outputSuccess` in `src/lib/output.ts`.

**Findings**
**Critical**
- `src/commands/login.ts:99-107`, `src/commands/login.ts:123-140`, `src/commands/login.ts:185-189`, `src/commands/config/plan.ts:69-70`, `src/commands/config/export.ts:78-79`: `--toon` is implemented by toggling module state that only `outputJson` / `outputTable` consult ([src/lib/output.ts](/tmp/codex-review-GcGMbL/CLI/src/lib/output.ts:13), [src/index.ts](/tmp/codex-review-GcGMbL/CLI/src/index.ts:159)). These command paths still call `console.log(JSON.stringify(...))` directly, so `insforge login --toon`, `insforge config plan --toon`, and `insforge config export --toon` will continue to emit JSON instead of TOON. That breaks the PR's core contract that `--toon` is a global structured-output format and also violates the repo's output convention in [DEVELOPMENT.md](/tmp/codex-review-GcGMbL/CLI/DEVELOPMENT.md:24).

**Suggestion**
- `src/lib/errors.ts:95-107`, `src/lib/output.test.ts:1-214`: error rendering still keys only off `json` and always prints JSON on structured error paths, so `--toon` failures will not be TOON-encoded. Separately, the new tests only exercise `src/lib/output.ts` in isolation and do not cover command-level `--toon` flows or error paths, which is why the direct `JSON.stringify` call sites above slipped through.

**Information**
- Software engineering: no additional style/convention issues beyond the direct-output bypass above.
- Security: no security-relevant changes beyond adding a pinned npm dependency in the lockfile.
- Performance: no material performance issues in the new TOON helpers; the added work is small relative to existing CLI rendering.

**Verdict**
`request_changes` because the global `--toon` behavior is not implemented consistently across existing structured command outputs.

jwfing (changes requested):
please address the latest comment。

PR #222, opened 2026-08-05
feat(local): add `insforge local start|stop|status`

Fermionic-Lyu:
Run a full InsForge backend on your own machine with one command — no account, no clone.

```bash
npx -y @insforge/cli@latest local start
```

Replaces "git clone, `cp .env.example .env`, `docker compose up`" as the self-host path. Pairs with InsForge/InsForge#1868, which makes the OSS image-only compose file drivable.

## How it works

The CLI bundles a copy of `deploy/docker-compose/docker-compose.yml`, which references only published images and named volumes — nothing is mounted from a checkout — so it runs correctly from inside the installed npm package. Ports, keys, and image refs go in through a generated `--env-file` rather than by rewriting YAML.

**One instance per directory.** The compose project name is derived from the directory path, so two app folders get separate containers, volumes, and databases with no configuration, and two directories sharing a basename don't collide.

**Zero-login by construction.** The backend seeds `ACCESS_API_KEY` / `ACCESS_ANON_KEY` on first boot, so the CLI picks the keys up front and links the directory without a round trip. `requireAuth()` already bypasses on the OSS sentinel project id, so `db` / `storage` / `functions` / `secrets` / `config` all target local immediately.

**Versioning.** The first start in a directory resolves the newest release present in *all three* InsForge image repos (same tag ⇒ same CI run ⇒ a combination that was tested together), pins each image by digest, and records that in `.insforge/local.json`. Later starts reuse it, so an existing project never moves on its own; new directories get the newest release automatically — no CLI publish needed per backend release.

`:latest` is deliberately not used. Docker won't re-pull a tag it already has, so it delivers neither reproducibility *nor* freshness, and the backend runs `migrate:up` on boot — a mismatched set of images can leave a schema an older image can't read. When the registry is unreachable, the versions bundled with the CLI are used; `local start` never fails on a network problem.

## Safety

- Secrets are generated once into `.insforge/local.env` at `0600`, with a scoped `.insforge/.gitignore` so `git add -A` can't commit them. Scoped deliberately — not `*`, which would start ignoring `project.json` for existing cloud-linked repos.
- A restart re-reads that file. Regenerating would rotate the API key away from the `.env.local` the app already holds.
- An existing cloud link is saved to `.insforge/project.cloud.json` rather than overwritten; `local stop --unlink` restores it. Mirrors what `branch switch` does with `project.parent.json`.
- `local stop --delete-data` is registered `critical` in the HITL guard — an agent running it wipes a database with no backup anywhere. A plain `stop` is `safe` and never prompts.
- Docker preflight distinguishes not-installed / daemon-down / compose-v1 / low-memory, and points at `insforge create` when there's no Docker at all.

## Verification

Ran against real containers, not just unit tests:

| Check | Result |
| --- | --- |
| Cold start → healthy | ~20s, all four containers up |
| `db query "select 1"`, `db tables` with no login | works |
| `storage create-bucket` / `buckets` / `upload` / `list-objects` | works |
| `local stop` then `local start` | same API key, same anon key, bucket survived |
| `--storage minio` + upload | object landed in `/data/insforge-storage/local/s3-bucket/` inside MinIO, not the filesystem |
| Cloud link displaced then `stop --unlink` | `project.cloud.json` written, then restored |
| `INSFORGE_GUARD=1 local stop --delete-data` | blocked, containers stayed up |
| `INSFORGE_GUARD=1 local stop` | ran, as intended |
| `local stop --delete-data` | no containers, no volumes, state and secrets cleared |
| `local status --json` | omits keys unless `--show-keys` |

Two bugs the tests caught while writing this, both worth noting because they'd have been silent:

- Store credentials were written under the overlay's variable names (`MINIO_ROOT_USER`) but read back under canonical ones, so `readSecrets` always returned `null` — every restart would have regenerated secrets and rotated the API key away from the app.
- `writeEnvFile` ran before `.insforge/` existed. My first test helper pre-created the directory and masked it; the helper now deliberately doesn't.

`npm run lint` (743 tests) and `npm run build` are clean; 51 new tests in `src/lib/local/`. Pre-existing `tsc --noEmit` errors elsewhere in the repo are untouched — none are in the new files.

## Follow-up

`postgres-all` and `deno-runtime` have no release tags yet (only `:latest`), so digest resolution correctly declines today and the bundled compose defaults are used — visible as "Using the versions bundled with this CLI". Once InsForge/InsForge#1868 merges and the next `v*` tag is cut, all three images exist at one tag and pinning activates with no change here.

---
## Summary by cubic
Run InsForge locally in Docker with one command. Adds `insforge local start|stop|status` to `@insforge/cli` for per-directory stacks, storage selection, port overrides, and zero-login linking.

- **New Features**
  - `local start` runs Postgres, PostgREST, backend, and Deno; waits for health; links this directory; seeds `.env.local`.
  - Storage: `--storage local|minio|rustfs` (S3 gateway for non-`local`). Ports: defaults 7130/7131/7133/5432/5430; override via `--port-*`; checks 0.0.0.0 and 127.0.0.1 to catch conflicts early.
  - Status/stop: `local status` shows health, ports, version, and per-service state (keys masked unless `--show-keys`, supports `--json`); `local stop` stops containers, keeps data by default.
  - Safety: secrets come from upstream setup and are re-used; cloud link is backed up and restorable via `local stop --unlink`; `local stop --delete-data` is HITL-guarded.
  - Uses upstream `setup.sh` and the unmodified compose file with a small CLI overlay setting `INSFORGE_DEPLOYMENT_METHOD=cli-local`; env and secrets live in `.insforge/checkout/.env`. Dropped image pinning and `--stack-tag`; version is read from `/api/health`.

- **Bug Fixes**
  - Prevent data loss on restart and deletion gaps: if `.insforge/checkout/.env` is missing but project volumes exist, `local start` now refuses with guidance; `local stop --delete-data` sweeps any leftover project volumes and reports what was removed.
  - Cross-platform and privacy: stable, Docker-legal compose project names on Windows (no whole-path names; same basename in different dirs stays separate); `--json` no longer includes existing `.env.local` values (only keys listed).

<sup>Written for commit 617228ebb35320355bb75f0725ea68154af90a4f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/CLI/pull/222?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Added `local start`, `local stop`, and `local status` commands for running InsForge locally with Docker.
  * Added configurable ports, storage options, stack versions, health checks, service status, and JSON output.
  * Added local persistence, secure credentials, cloud-link backup and restoration, and optional data deletion.
  * Added Docker readiness checks and automatic retrieval and caching of required local assets.

* **Documentation**
  * Added comprehensive guidance for setup, configuration, status inspection, shutdown, data management, and troubleshooting.

* **Tests**
  * Added extensive coverage for local deployment, ports, configuration rendering, secrets, state management, registries, and upstream assets.

> [!NOTE]
> ### Add `insforge local start|stop|status` commands to manage a local Docker Compose stack
> - Adds three new CLI subcommands under `local`: `start`, `stop`, and `status`, registered in [src/commands/local/](https://github.com/InsForge/CLI/pull/222/files#diff-92b509503c8a8bdb17edc13b65977b648406c3c2090e85eccc31c06f74d666d1).
> - `local start` provisions a multi-container stack (postgres, postgrest, insforge, deno) via Docker Compose, generates and persists secrets, resolves a stack image tag from GHCR, fetches upstream config files, waits for backend health, and writes connection details to `.env.local`.
> - `local stop` tears down the stack; `--delete-data` removes volumes and forgets local state; `--unlink` restores a previously backed-up cloud project link.
> - `local status` reports container states, backend health, URLs, and optionally reveals API keys via `--show-keys`.
> - State, secrets, and compose files are written under `.insforge/` with a managed `.gitignore` to prevent accidental commits.
> - The bundled [docker-compose.template.yml](https://github.com/InsForge/CLI/pull/222/files#diff-86f4c66e0863bfeb2ce2c0168449665c6424f836c3818b5f31e13948541c0041) is shipped in the package and rendered at runtime with inlined config payloads.
> - Risk: `local stop --delete-data` is classified as critical (irreversible volume deletion) in the guard risk registry; plain `local stop` is safe.
>
> 
> #### Changes since #222 opened
>
> - Introduced upstream stack provisioning via setup.sh fetch and execution [63041a9]
> - Replaced CLI-generated secrets and environment management with setup.sh-generated secrets and delta-based updates [63041a9]
> - Refactored Docker Compose configuration to use upstream compose file with CLI overlay instead of rendered templates [63041a9]
> - Removed stackTag tracking and replaced with version from backend health endpoint across local commands [63041a9]
> - Updated build configuration to ship cli-overlay.yml instead of docker-compose.template.yml [63041a9]
> - Modified `ensureCheckout` to accept an optional `onExistingData` callback and abort with a `CLIError` when the `.env` file is missing but existing project volumes are detected, and integrated this protection into `registerLocalStartCommand` by passing `projectVolumes` as the callback [686767e]
> - Added `projectVolumes` and `removeProjectVolumes` utilities to list and remove Docker volumes owned by a compose project, and integrated volume removal into `registerLocalStopCommand` when `--delete-data` is supplied [686767e]
> - Added test coverage for `ensureCheckout` behavior with existing volumes and missing `.env` file [686767e]
> - Modified JSON output format for `local start` command success payload [617228e]
> - Fixed cross-platform path handling in `composeProjectName` function and added test coverage [617228e]
> 
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 9e8a945.</sup>
> 
>


jwfing (commented):
## Code review — `feat(local): add \`insforge local start|stop|status\``

**Summary:** A well-structured, well-tested feature that turns self-hosting into a single command; the extracted lib helpers have solid unit coverage and the security posture (0600 secrets, scoped `.gitignore`, HITL guard on data deletion, masked keys, no secret sent to PostHog) is thought through — no blocking issues, three non-blocking suggestions below.

### Requirements context
No matching spec/plan found — `docs/specs/` contains only the `diagnose` and `db-migrations` designs, nothing for the `local` command. Assessed against the PR description, `DEVELOPMENT.md`, and the paired `InsForge/InsForge#1868` (image-only compose). Conventions in `DEVELOPMENT.md` were used as the standards baseline.

---

### Critical
(none)

---

### Suggestions

**1. Functionality / robustness — a truncated `local.env` silently rotates the API key, contradicting the module's own contract.** `src/lib/local/secrets.ts:82-88` deliberately returns `null` on a *partial* file and documents *"let the caller refuse rather than silently half-configure compose."* But the only caller, `src/commands/local/start.ts:171`, does `const secrets = readSecrets() ?? generateSecrets();` — it does **not** refuse. A truncated env file (partial write / disk-full / hand-edit) therefore takes the same branch as a first start: new secrets are generated and written, while the running Postgres volume still holds the *old* seeded `ACCESS_API_KEY`. The result is `project.json`/`.env.local` pointing at a key the backend rejects — exactly the "rotate the key out from under the app" failure the whole secrets design exists to prevent. It's an uncommon trigger and recoverable via `local stop --delete-data`, hence a suggestion rather than a blocker, but the caller should distinguish *absent* (`!existsSync` → generate, correct on first start) from *present-but-unreadable* (throw a `CLIError` telling the user how to recover). Consider having `readSecrets` throw on a partial file instead of collapsing both cases to `null`. The `returns null on a truncated file` test at `src/lib/local/secrets.test.ts:117` pins the current behavior but nothing asserts what `start` does with it.

**2. Security — the app port binds to all interfaces while every sibling port is loopback-scoped.** In `src/assets/local/docker-compose.yml`, postgres (`:26`), postgrest (`:49`), auth (`:73`) and deno (`:163`) all publish as `127.0.0.1:<port>:...`, but the app/dashboard on `:72` is `"${APP_PORT:-7130}:7130"` — i.e. `0.0.0.0`, reachable from the LAN. For a tool explicitly framed as "run on your own machine," that exposes the dashboard and any unauthenticated API surface to anyone on the same network. Privileged operations still need the (random) API key and admin password, so this isn't an open-admin hole — hence a suggestion. If cross-device access (e.g. testing from a phone) is the intent, that's reasonable but should be documented; otherwise bind `127.0.0.1:${APP_PORT:-7130}:7130` for consistency with the other four services. Worth a one-line decision either way, since the asymmetry reads as an oversight.

**3. Software engineering — command orchestration and the `__testing` exports have no tests.** The lib layer (`compose`/`ports`/`registry`/`secrets`/`state`) is covered well by the 51 new tests. But the three action handlers (`start.ts`/`stop.ts`/`status.ts`) have no test files, and `src/commands/local/start.ts:314` exports `__testing = { parsePort, portOverrides, resolveStorage }` "for tests" that don't exist yet. The orchestration logic that isn't trivially container-dependent — the cloud-link backup/restore branching (`start.ts:102-110`, `stop.ts:19-25,60-68`) and `parsePort`/`resolveStorage` validation — is unit-testable without Docker and is where a close-call regression (e.g. clobbering a cloud `project.json`) would bite. Manual verification per the PR table is great; a few pure-function tests would lock it in.

---

### Information

- **Fallback stack pins `:latest` for two of three images.** `src/assets/local/docker-compose.yml:15,156` default postgres-all/deno-runtime to `${INSFORGE_STACK_TAG:-latest}` while `insforge` defaults to `v2.2.9`. This is the exact non-reproducibility/non-freshness footgun `src/lib/local/registry.ts:1-18` documents, taken deliberately as the interim fallback until `InsForge/InsForge#1868` cuts a common `v*` tag. Fine as described — flagging only so it isn't forgotten once tags exist.
- **`--stack-tag` is written to `local.env`/`local.json` unvalidated.** The resolved-tag path is regex-gated by `RELEASE_TAG` (`registry.ts:37`), but an explicit `--stack-tag` value flows straight into `renderEnvFile` (`secrets.ts:133-134`) and `LocalState` without a shape check. Self-inflicted only (the user's own machine/compose), so not a security issue, but validating against `/^v\d+\.\d+\.\d+$/` would give a cleaner error than a malformed env file.
- **Agent-skill sync (`DEVELOPMENT.md` §3).** Adding a new user-facing command should be mirrored in the `insforge-cli` skill in `InsForge/agent-skills`, and the PR body should cross-reference that PR. Worth confirming a companion update exists so agents learn `local`.
- **Analytics convention satisfied.** `trackCommandUsage` (used by `storage`/`secrets`/`schedules`) is the established wrapper and only non-sensitive props (`storage`, `stack_tag`, `delete_data`, `healthy`) are sent — compliant with `DEVELOPMENT.md` §2. `package.json` `files` correctly whitelists the three bundled compose assets, and `tsup.config.ts` copies them to `dist/assets/local/`, so the npm package will ship them.

**Performance:** no concerns — registry resolution is bounded by an 8s abort and runs only on first start; the health poll is a reasonable 2s interval. **Nice touches:** the paginated GHCR tag listing (`registry.ts:90-110`) correctly avoids the lexicographic-first-page trap, `composeProjectName` hashing the absolute path prevents same-basename collisions, and the escalate-only HITL entry for `local stop --delete-data` is exactly right.

---

### Verdict
**approved** (informational — no Critical findings; the three items above are non-blocking). Human GitHub approval remains a separate action.

jwfing (dismissed):
LGTM - approved.

Fermionic-Lyu:
Superseded by https://github.com/InsForge/CLI/pull/226 — same feature, rebased on current `main`, with the superseded approaches squashed out into three reviewable commits.

PR #227, opened 2026-08-08
feat(compute): support the self-hosted Docker provider

Fermionic-Lyu:
InsForge/InsForge#1892 added a self-hosted Docker compute driver. `compute deploy` still assumed Fly, so a self-hoster on the new driver got a region silently recorded and never honoured — and `compute deploy <dir>` could not work at all, since source mode minted a Fly deploy token and shelled out to `flyctl`.

The CLI now asks `GET /api/metadata` what the configured provider can do and shapes the request accordingly.

## What changed

- **`region`** is omitted when the provider has none, so the stored row doesn't claim a choice that never took effect. Passing `--region` explicitly prints that it's being ignored rather than failing.
- **`scaleToZero`** likewise, including `--always-on`.
- **Source mode branches on `sourceBuild`:**
  - `context-upload` (Docker) packs the directory and POSTs it to `/:id/build`, which builds, tags and deploys in one call — no deploy token, no `flyctl` on PATH, no follow-up PATCH.
  - `flyctl` (Fly) keeps the existing remote-build path unchanged.
  - `none` fails with a message naming `--image` as the way through. Checked *before* the Dockerfile lookup, since a missing Dockerfile is beside the point when the provider can't build regardless.

## The context packer

Deliberately **not** modelled on the deployments bundler, which strips `node_modules`, `dist`, `build`. Those exclusions are right for a static site and wrong for a Docker build: a Dockerfile may legitimately `COPY` any of them, and silently dropping one produces a build failure that nothing on screen explains.

It follows Docker's own contract — everything, minus `.dockerignore` — with `.git` the single unconditional exclusion, since it's usually the largest thing in the tree and carries every secret ever committed. When a context is too big, the backend already answers 413 and names `.dockerignore` as the fix.

Built with `archiver` and `ignore`, both already dependencies here, so **no new dependency**.

## Compatibility

The capability lookup never throws. A CLI that can't deploy because a probe failed is worse than one that tries the way it always has, so any failure — and any backend older than the `compute` slice — falls back to the Fly-shaped behaviour, field by field.

## Verification

782 tests pass; eslint clean.

Existing deploy tests asserted on `ossFetch` call *indices*, which the capability probe shifted by one. They now route the mock by URL, which also survives the next call anyone adds. Eight new tests: four on capability-driven request shaping (region omitted / still sent, scaleToZero omitted, `sourceBuild: none` refused) and four on the packer (default inclusion, `.dockerignore` with negations, `.git` exclusion, tar block validity). Both new behaviours verified by reverting the code and watching the tests fail.

Note: this repo has no prettier config, so the diff is hand-matched to the surrounding single-quote style — no reformatting noise.

---
## Summary by cubic
Add self-hosted Docker provider support to `compute deploy`. The CLI now reads `/api/metadata` capabilities and adapts deploy requests and source builds accordingly.

- New Features
  - Probe capabilities via `/api/metadata`; fall back to Fly-shaped defaults if missing.
  - Skip `region` when the provider has none; show a note if `--region` is passed; still send it for multi-region providers.
  - Skip `scaleToZero` when unsupported; ignore `--always-on` accordingly.
  - Source builds respect `sourceBuild`: `context-upload` packs the directory and POSTs to `/:id/build` (no deploy token, no `flyctl`), `flyctl` keeps the current path, `none` errors with guidance to use `--image`.
  - New build-context packer follows `.dockerignore` and always excludes `.git`; uses `archiver` and `ignore` (no new deps).

- Refactors
  - Tests route `ossFetch` mocks by URL instead of call order; added coverage for capability-driven shaping and context packing.

<sup>Written for commit 62ef0de16e4944b555b9e9cc13edeb9f62372120. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/CLI/pull/227?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

> [!NOTE]
> ### Add self-hosted Docker provider support to the compute deploy command
> - Adds [`fetchComputeCapabilities`](https://github.com/InsForge/CLI/pull/227/files#diff-8e9591bf04723aeff6e42472816859c3f6d8c5444a21c6ea27e44c2000c77ab1) to query `/api/metadata` for provider capabilities, with fallbacks to Fly.io defaults on missing or failed responses.
> - The deploy command now conditionally omits `region` and `scaleToZero` from request bodies when the provider does not support them, logging an info message in each case.
> - Adds a new `context-upload` source mode via [`packBuildContext`](https://github.com/InsForge/CLI/pull/227/files#diff-1ecbda791e7f463961036fdfef1255294d3e1775d1528f81f241e5d9d2452293), which tars the build directory (honoring `.dockerignore`, always excluding `.git`) and uploads it to the backend build endpoint.
> - Throws a `CLIError` when source mode is requested but the provider has `sourceBuild: 'none'`, directing users to build an image manually.
> - `flyctl` is now only required for providers that use the flyctl-based build path.
> - Risk: providers without region or scale-to-zero support silently drop those fields rather than erroring, which may be unexpected if the flag was intentional.
>
> 
>
> <sup><a href="https://app.macroscope.com">Macroscope</a> summarized 62ef0de.</sup>
> 
>

### DIFF
--- a/src/commands/compute/deploy.test.ts
+++ b/src/commands/compute/deploy.test.ts
@@ -16,13 +16,65 @@ vi.mock('../../lib/errors.js', async (importOriginal) => {
 import { Command } from 'commander';
 import { registerComputeDeployCommand } from './deploy.js';
 
+/**
+ * Route the mock by URL rather than by call order.
+ *
+ * `compute deploy` now asks /api/metadata what the provider can do before shaping
+ * the request, so an assertion keyed on `mock.calls[1]` breaks as soon as another
+ * call is added. Routing by URL survives that.
+ */
+function routeOssFetch(capabilities?: Record<string, unknown>) {
+  const caps =
+    capabilities ?? {
+      scaleToZero: true,
+      regions: true,
+      ingressModes: ['host'],
+      sourceBuild: 'flyctl',
+      deployTokenIssuance: true,
+    };
+  ossFetchMock.mockImplementation((url: string, init?: { method?: string }) => {
+    if (url === '/api/metadata') {
+      return Promise.resolve({
+        json: async () => ({ compute: { defaultProvider: 'test', providers: { test: caps } } }),
+      });
+    }
+    if (url === '/api/compute/services' && !init?.method) {
+      return Promise.resolve({ json: async () => [] });
+    }
+    return Promise.resolve({
+      json: async () => ({
+        id: 'svc-1',
+        name: 'cache',
+        status: 'started',
+        endpointUrl: 'https://cache.fly.dev',
+        port: 6379,
+        service: { name: 'cache', status: 'running' },
+        imageTag: 'insforge-x/cache:abc',
+        logs: ['Step 1/2 : FROM alpine'],
+      }),
+    });
+  });
+}
+
+/** Body of the POST that creates or prepares the service. */
+function createCallBody(): Record<string, unknown> {
+  const call = ossFetchMock.mock.calls.find(
+    ([url, init]) =>
+      typeof url === 'string' &&
+      url.startsWith('/api/compute/services') &&
+      (init as { method?: string } | undefined)?.method === 'POST'
+  );
+  if (!call) {
+    throw new Error('no create/prepare POST was made');
+  }
+  return JSON.parse((call[1] as { body: string }).body) as Record<string, unknown>;
+}
+
+
 describe('compute deploy --protocol', () => {
   beforeEach(() => {
     ossFetchMock.mockReset();
-    ossFetchMock.mockResolvedValueOnce({ json: async () => [] }); // initial list
-    ossFetchMock.mockResolvedValueOnce({
-      json: async () => ({ name: 'cache', status: 'started', endpointUrl: 'https://cache.fly.dev', port: 6379 }),
-    });
+    routeOssFetch();
   });
 
   it('includes protocol="tcp" in request body when --protocol tcp', async () => {
@@ -37,8 +89,7 @@ describe('compute deploy --protocol', () => {
       '--protocol', 'tcp',
       '--port', '6379',
     ]);
-    const createCall = ossFetchMock.mock.calls[1];
-    const body = JSON.parse(createCall[1].body);
+    const body = createCallBody();
     expect(body.protocol).toBe('tcp');
     expect(body.port).toBe(6379);
   });
@@ -52,8 +103,7 @@ describe('compute deploy --protocol', () => {
       'node', 'lim', 'compute', 'deploy',
       '--image', 'nginx', '--name', 'web', '--port', '8080',
     ]);
-    const createCall = ossFetchMock.mock.calls[1];
-    const body = JSON.parse(createCall[1].body);
+    const body = createCallBody();
     expect('protocol' in body).toBe(false);
   });
 
@@ -74,10 +124,7 @@ describe('compute deploy --protocol', () => {
 describe('compute deploy --always-on / --scale-to-zero', () => {
   beforeEach(() => {
     ossFetchMock.mockReset();
-    ossFetchMock.mockResolvedValueOnce({ json: async () => [] }); // initial list
-    ossFetchMock.mockResolvedValueOnce({
-      json: async () => ({ name: 'api', status: 'running', endpointUrl: 'https://api.fly.dev', port: 8080, scaleToZero: false }),
-    });
+    routeOssFetch();
   });
 
   it('includes scaleToZero=false in request body when --always-on', async () => {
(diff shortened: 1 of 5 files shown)


jwfing (commented):
## Review: feat(compute): support the self-hosted Docker provider

**Summary:** A well-structured, defensively-designed change that teaches `compute deploy` to probe `/api/metadata` for provider capabilities and route source builds through a context upload for the self-hosted Docker driver, with a Fly-shaped fallback that never throws — no Critical issues, a few worthwhile follow-ups.

### Requirements context
No matching spec/plan found under `docs/specs/` (only `2026-03-27-diagnose-*` and `2026-04-17-db-migrations-*` live there; this repo has no `docs/superpowers/`). Assessed against the PR description and the referenced backend PR `InsForge/InsForge#1892`. I verified the two dependency claims (`archiver@^7`, `ignore@^7` are already runtime deps used in `src/commands/deployments/`), and that `ossFetch` spreads `options` so the `Buffer` tar body reaches `fetch` unencoded with the `application/x-tar` header overriding the default `application/json` (`src/lib/api/oss.ts:240-246`) — the `/build` POST is wired correctly.

### Findings

#### Critical
(none)

#### Suggestion

- **[software eng / analytics] Context-upload success path skips PostHog telemetry.** `src/commands/compute/deploy.ts:308` calls only `reportCliUsage(...)` on success, but both other success paths — image mode (`:191`) and flyctl source mode (`:419`) — also call `await trackCommandUsage('compute', 'deploy', true)`. `command-telemetry.ts` states every command should emit *exactly one* PostHog event per invocation, and `DEVELOPMENT.md §2` names PostHog the source of truth going forward. As written, a successful self-hosted Docker deploy emits **no** success event (only the `catch` at `:442` fires `trackCommandUsage` on failure), so those deploys are undercounted. Add the `trackCommandUsage('compute', 'deploy', true)` call to the context-upload path.

- **[software eng / test coverage] The context-upload happy path and rollback are untested.** The eight new tests cover capability-driven request *shaping* (region/scaleToZero omitted or kept, `sourceBuild: 'none'` refused) and the packer, but nothing drives a `deploy <dir>` through `sourceBuild: 'context-upload'` — the list→prepare→pack→`POST /:id/build`→output flow (`deploy.ts:239-309`) and the build-failure rollback (`:275-285`) ship with no coverage. That's the core new behavior; a regression there (including the missing telemetry call above) wouldn't be caught. Consider a test that routes the mocked `/build` and asserts the tar upload + rollback-on-failure.

- **[functionality] Symlinks are silently dropped from the build context.** `src/lib/build-context.ts:74-77` skips anything that isn't `entry.isFile()` (or a directory), so a symlink — even a valid one pointing inside the context — is excluded. This contradicts the module's own stated "everything, minus `.dockerignore`" contract, and reproduces the exact footgun the packer was written to avoid: a Dockerfile that `COPY`s a symlinked file/dir gets a build failure with nothing on screen explaining it. At minimum, follow the symlink for in-tree targets or note the exclusion.

- **[functionality / UX] `--region` notice uses `process.argv` instead of the parsed args.** `deploy.ts:136` gates the "Ignoring --region" message on `process.argv.includes('--region')`. Because `--region` has a default (`:49`), you can't read intent from `opts` — but the `process.argv` scan misses the `--region=lhr` equals form and is decoupled from what Commander actually parsed (which is also why the message can't be exercised via `parseAsync`). `cmd.getOptionValueSource('region') === 'cli'` is the robust, testable check. Note this only affects the advisory message; the actual region omission is correctly driven by `capabilities.regions`.

#### Information

- **[security] Default context includes `.env`/secrets unless `.dockerignore`'d.** This is intentional and matches Docker's own contract (and the context goes to the user's own self-hosted backend), but it differs from the static deployments bundler users may expect — worth a line in `--help`/docs so no one is surprised that `.env` ships by default. `.git` is correctly excluded unconditionally.
- **[functionality] `--scale-to-zero` on a provider without scale-to-zero is dropped without a notice** — only `--always-on` (`scaleToZero === false`) prints the "Ignoring" message (`deploy.ts:141`). Minor asymmetry.
- The capability fallback design is sound: `fetchComputeCapabilities` never throws and degrades field-by-field to `LEGACY_FLY`, so an old backend or a failed probe keeps the historical Fly behavior. `serviceId` is `encodeURIComponent`'d on every path. **No security-relevant regressions.** **Performance:** the in-memory tar is bounded by the backend's 64MB ceiling and only one extra `/api/metadata` round-trip is added — acceptable for a CLI.

### Verdict
**approved** — no Critical findings. The Suggestions (missing success telemetry + untested happy path especially) are worth addressing before merge but are non-blocking. (Informational verdict; explicit GitHub approval remains a separate human action.)

jwfing (approved):
LGTM - approved.

PR #267, opened 2026-08-17
fix(branch): survive a transient control-plane 502 while polling (v0.2.8)

Fermionic-Lyu:
## The bug

`insforge branch create` polls `GET /projects/v1/branches/:id` every 3s for up to 15 minutes. A **single** 502 anywhere in that window ended the command on the spot — while the backend went on to mark the branch ready seconds later. The branch exists and bills; the CLI exited non-zero with no id in the output.

Reported from agent-e2e, and confirmed in the code — [`pollUntilReady`](https://github.com/InsForge/CLI/blob/191cac6/src/commands/branch/create.ts#L277) called `getBranchApi` bare, and `platformFetch` throws on any non-ok response with no retry.

| run | CLI | backend |
|---|---|---|
| `31832239687` | 502, exit at ~90s | ready at ~108s |
| `32055449431` | 502, exit at ~97s | ready at ~113s |
| `32005660955` (green) | — | ready at ~113s |

The duration itself is normal — the third row is a successful run taking the same ~113s. This is a CLI resilience bug, not an agent-e2e retry problem.

Same bug existed in `branch reset` (identical poll, same endpoint). `waitUntilServing` was already tolerant, and `deployments deploy` already had exactly this tolerance — the branch pollers were the odd ones out.

## The fix

A failed **read** is not a failed branch.

- **`isTransientApiError`** (`lib/errors.ts`) — gateway 5xx, 408/429, and tagged network failures are transient; every other 4xx and any locally raised `CLIError` (e.g. a terminal `deleted`/`conflicted` state) stays terminal, so a real answer still ends the loop on the first call. `deployments/deploy.ts` carried an inline copy of this rule and now shares the helper.
- **`platformFetch` attaches the HTTP status** to the `CLIError` it throws, so a poller can tell a 502 from a 404 at all. `CLIError.statusCode` already existed and was already read by command telemetry — only `ossFetch` had been populating it. No existing `statusCode` check reads platform errors, so nothing else changes behavior.
- **Both branch pollers wrap only the fetch**: a transient failure costs one poll interval and nothing more. The post-timeout re-check falls back to the last observed state rather than turning a timeout into an API error about a branch that exists.

### Also: the same hole on the create `POST`

A 502 there is worse — it orphans a billing branch with no id anywhere in the output, which is the exact harm the existing adopt path was built to prevent. A gateway 5xx is the proxy reporting that *it* could not complete the round trip, so it now joins transport resets as an ambiguous failure eligible for adoption. Guards are unchanged (name + mode + `created_at >= requestedAt`), and an unmatched search still rethrows the original error — so this cannot mask a request the backend never acted on.

## Tests

- regression test reproducing the incident: one 502 mid-poll → still exits 0 with the ready branch
- the mirror case: a 404 mid-poll is still terminal after exactly one call
- 5xx adoption on create; both cases repeated for `branch reset`
- unit coverage of the predicate across 5xx / 408 / 429 / 4xx / no-status / non-`CLIError`

`npm run lint` (vitest + eslint): **792 passed**, 13 skipped, eslint clean.

## Not covered

A *sustained* control-plane outage across the whole 15-minute budget still ends in "creation failed" without echoing the branch id. Narrowing that would mean distinguishing it from a genuine `deleted`/`conflicted` failure, whose message is accurate today — left for a follow-up.

No `insforge-cli` skill update needed: no flags, commands, or output shapes changed.

Bump 0.2.7 → 0.2.8.

---
## Summary by cubic
Tolerates transient failures while polling branches and aligns deployment polling with the same rules. Previously, a single 502 on `GET /projects/v1/branches/:id` aborted `insforge branch create`/`branch reset`, and deployment polls retried unclassified errors for the whole window; now pollers retry only classified-transient reads and fail on real rejections.

- Adds `isTransientApiError`: all 5xx, 408/429, and tagged network failures are transient; other 4xx, locally raised `CLIError`s, and non-`CLIError` throws are terminal.
- `platformFetch` now sets `CLIError.statusCode`; `ossFetch` now tags transport failures as `CLIError(NETWORK_ERROR_CODE)` with an actionable message.
- `branch create`/`reset` pollers wrap only the read; transient failures cost one interval and continue. The final verdict read is retried; on timeout, create falls back to the last observed state to emit id/appkey, while reset fails loudly if the final state cannot be confirmed.
- Create `POST`: adoption is limited to ambiguous failures (transport resets and proxy 502/503/504). Plain 500 and 408/429 rethrow unchanged. Guards remain name, mode, and `created_at >= request time`.
- `deployments deploy` poller switches to `isTransientApiError`; malformed status bodies and other non-transient errors now fail fast instead of being retried for the full window.
- Version bump to 0.2.8.

**Review and rollout**
- Focus on the transient classifier, `ossFetch` transport tagging, final-read retry helper, and the narrower create‑adoption predicate.
- No flags or output formats changed; no migration required.

<sup>Written for commit fbffb747845227154ea2b6b058dded41e5cc68e4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/InsForge/CLI/pull/267?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@insforge/cli",
-  "version": "0.2.7",
+  "version": "0.2.8",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@insforge/cli",
-      "version": "0.2.7",
+      "version": "0.2.8",
       "license": "Apache-2.0",
       "dependencies": {
         "@clack/prompts": "^0.9.1",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@insforge/cli",
-  "version": "0.2.7",
+  "version": "0.2.8",
   "description": "InsForge CLI - Command line tool for InsForge platform",
   "type": "module",
   "bin": {
--- a/src/commands/branch/create.test.ts
+++ b/src/commands/branch/create.test.ts
@@ -394,6 +394,118 @@ describe('branch create', () => {
     expect(exitCode).toBe(1);
   });
 
+  it('survives a transient 502 mid-poll instead of abandoning a branch that is still provisioning', async () => {
+    // The reported incident: the control plane 502s once at ~90s, the CLI exits
+    // non-zero, and the backend marks the branch ready ~15s later — leaving a
+    // real, billing branch behind a failed command. A failed READ is not a
+    // failed branch. (agent-e2e runs 31832239687 / 32055449431.)
+    const { getProjectConfig } = await import('../../lib/config.js');
+    (getProjectConfig as Mock).mockReturnValue({
+      project_id: 'p1',
+      project_name: 'parent',
+      org_id: 'o1',
+      appkey: 'p1ky',
+      region: 'us-east',
+      api_key: 'k',
+      oss_host: 'p1ky.us-east.insforge.app',
+    });
+    const { getBranchApi } = await import('../../lib/api/platform.js');
+    // One gateway 502, then the branch is ready (the default mock impl).
+    (getBranchApi as Mock).mockRejectedValueOnce(
+      new CLIError('Request failed: 502', 1, undefined, 502),
+    );
+    const logs: string[] = [];
+    const origLog = console.log;
+    console.log = (...args: unknown[]) => {
+      logs.push(args.map(String).join(' '));
+    };
+    vi.useFakeTimers();
+    let exitCode: number | undefined;
+    try {
+      exitCode = await withCapturedExit(async () => {
+        const program = new Command().exitOverride();
+        program.option('--json').option('--api-url <url>').option('-y, --yes');
+        registerBranchCreateCommand(program);
+        const run = program
+          .parseAsync(['create', 'feat-x', '--mode', 'schema-only', '--no-switch', '--json'], {
+            from: 'user',
+          })
+          .catch(() => {});
+        await vi.runAllTimersAsync();
+        await run;
+      });
+    } finally {
+      vi.useRealTimers();
+      console.log = origLog;
+    }
+    // Polled past the 502 and reported the ready branch, exit 0.
+    expect((getBranchApi as Mock).mock.calls.length).toBeGreaterThan(1);
+    expect(exitCode).toBeUndefined();
+    expect(logs.join('\n')).toContain('branch-id');
+  });
+
+  it('still gives up on a real rejection mid-poll (404 is not transient)', async () => {
+    // The tolerance must not swallow an answer that will never change — a
+    // deleted/unknown branch id has to end the command, not burn 15 minutes.
+    const { getProjectConfig } = await import('../../lib/config.js');
+    (getProjectConfig as Mock).mockReturnValue({
+      project_id: 'p1',
+      project_name: 'parent',
+      org_id: 'o1',
+    });
+    const { getBranchApi } = await import('../../lib/api/platform.js');
+    (getBranchApi as Mock).mockRejectedValueOnce(
+      new CLIError('Branch not found', 1, undefined, 404),
+    );
+    const exitCode = await withCapturedExit(async () => {
+      const program = new Command().exitOverride();
+      program.option('--json').option('--api-url <url>').option('-y, --yes');
+      registerBranchCreateCommand(program);
+      await program
+        .parseAsync(['create', 'feat-x', '--mode', 'schema-only', '--no-switch'], { from: 'user' })
+        .catch(() => {});
+    });
+    expect(getBranchApi as Mock).toHaveBeenCalledTimes(1);
+    expect(exitCode).toBe(1);
+  });
+
+  it('adopts a branch that was created despite a gateway 5xx on the create request', async () => {
+    // Same lost-response ambiguity as a transport reset: a 502 is the proxy
+    // saying IT could not complete the round trip, so the branch may well exist
+    // and be billing. The name/mode/created_at guards still apply.
+    const { getProjectConfig } = await import('../../lib/config.js');
+    (getProjectConfig as Mock).mockReturnValue({
+      project_id: 'p1',
+      project_name: 'parent',
+      org_id: 'o1',
+    });
+    const { createBranchApi, listBranchesApi } = await import('../../lib/api/platform.js');
+    (createBranchApi as Mock).mockRejectedValueOnce(
+      new CLIError('Request failed: 502', 1, undefined, 502),
+    );
+    (listBranchesApi as Mock).mockResolvedValueOnce([
+      {
+        id: 'branch-id',
+        parent_project_id: 'p1',
+        organization_id: 'o1',
+        name: 'feat-x',
+        appkey: 'p1ky-x9p',
+        region: 'us-east',
+        branch_state: 'creating',
+        branch_created_at: new Date().toISOString(),
+        branch_metadata: { mode: 'schema-only' },
+      },
+    ]);
+    const program = new Command().exitOverride();
+    program.option('--json').option('--api-url <url>').option('-y, --yes');
+    registerBranchCreateCommand(program);
+    await program
+      .parseAsync(['create', 'feat-x', '--mode', 'schema-only', '--no-switch'], { from: 'user' })
+      .catch(() => {});
+    expect(listBranchesApi as Mock).toHaveBeenCalledWith('p1', undefined);
+    expect(String(spinnerMock.stop.mock.calls.at(-1)?.[0])).not.toContain('creation failed');
+  });
+
   it('adopts a branch that was created despite a transport failure', async () => {
     // createBranchApi carries no idempotency key, so a reset on the RESPONSE
     // leg leaves a real, billing branch behind. Giving up here orphans it.
--- a/src/commands/branch/create.test.ts
+++ b/src/commands/branch/create.test.ts
@@ -506,6 +506,33 @@ describe('branch create', () => {
     expect(String(spinnerMock.stop.mock.calls.at(-1)?.[0])).not.toContain('creation failed');
   });
 
+  it('does NOT adopt on a plain 500 — that is the API answering, not a lost response', async () => {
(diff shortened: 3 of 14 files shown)


jwfing (commented):
## Review: fix(branch): survive a transient control-plane 502 while polling (v0.2.8)

**Summary:** A tightly-scoped, well-tested resilience fix — a failed *read* during branch-poll no longer aborts a branch that is still provisioning — with a clean helper extraction that de-duplicates the pre-existing deploy tolerance and, importantly, preserves deploy's behavior exactly.

### Requirements context
No matching spec/plan found. The repo has `docs/specs/` (covering only `diagnose` and `db-migrations` command designs) and no `docs/superpowers/`; nothing there addresses branch polling. Assessed against the PR description, the linked incident (agent-e2e runs `31832239687` / `32055449431`), and `InsForge/InsForge#1790`.

### Findings

#### Critical
(none)

#### Suggestion
- **[software-eng / UX] The two branch pollers diverge in how they re-announce after recovery** — `create.ts:297-322` parks an `UNREACHABLE_STATE` sentinel in `lastState` so the next successful read *re-announces* the true state even if unchanged, whereas `reset.ts:99-121` uses a separate `announcedUnreachable` boolean and leaves `lastState` untouched, so it does **not** re-announce the state on recovery. Both are correct; the spinner/`outputInfo` copy is just slightly inconsistent between the two commands. Purely cosmetic — worth aligning if you touch this again, not worth blocking.

#### Information
- **[functionality] Non-`CLIError` throws are transient — a slight widening of the silent-retry window in the branch pollers.** `isTransientApiError` (`src/lib/errors.ts:66-71`) returns `true` for any non-`CLIError`. The branch pollers previously had *zero* tolerance (`getBranchApi` was called bare), so a genuine client-side bug that throws a raw error (e.g. a `res.json()` `SyntaxError` on a truncated proxy body, or a real programming error) would previously surface immediately and now gets retried silently for up to the full 15-minute budget. In practice this is the intended trade-off — a broken-gateway body *should* be retried, and `platformFetch` wraps all HTTP/network failures into `CLIError` — so the only thing masked is a true bug. Documented in the function's own comment; noting it only so the widened surface is a conscious choice.
- **[functionality] `>= 500` treats permanent server errors (501/505) as transient**, burning the whole poll budget before "still in state …". These statuses don't occur for these control-plane endpoints, and the prior deploy logic did the same, so this is a non-issue — flagged only for completeness.

### Verification performed
- **Deploy refactor is behavior-preserving.** Confirmed `ossFetch` (`src/lib/api/oss.ts:301`) always throws `CLIError(msg, 1, err.error, res.status)` with `statusCode` defined and `code` set to the backend error string (never `NETWORK_ERROR_CODE`), and that raw fetch failures propagate un-wrapped as non-`CLIError`. Mapping the old inline `isTerminal` predicate onto `!isTransientApiError` yields identical outcomes for every case deploy can produce (defined-statusCode CLIErrors, locally-raised undefined-statusCode CLIErrors, and raw throws). `TRANSIENT_4XX_STATUSES = {408, 429}` carried over unchanged.
- **`platformFetch` now attaches status on both throw paths** (`src/lib/api/platform.ts:124-129` retry-leg, `:144` main-leg), so pollers can distinguish 502 from 404. `CLIError`'s 4th ctor arg (`statusCode`, `src/lib/errors.ts:19`) already existed and was already read by telemetry — no ripple.
- **Adopt path stays safe.** `isAmbiguousCreateFailure` (`create.ts:224-228`) is correctly *narrower* than the poll predicate — network reset or 5xx only, **not** 408/429 (a rate-limit on the POST means the branch was *not* created, so adopting would be wrong). The three guards (name + mode + `created_at >= requestedAt`) are unchanged, the `listBranchesApi(...).catch(() => undefined)` swallows a failed lookup into a rethrow of the original error, and an unmatched search rethrows — so 5xx widening cannot mask a request the backend never acted on (`create.ts:238-252`).
- **Auth stays terminal.** A 401 triggers `platformFetch`'s refresh-and-retry; a persistent 401 throws with `statusCode=401` → `isTransientApiError` false → loop ends. No auth weakening.
- **Timeout fallback.** Final re-check falls back to `lastBranch` only on a transient failure *and* only when a prior read succeeded (`create.ts:329-332`, `reset.ts:135-140`); a from-the-start outage with `lastBranch === null` still rethrows — matching the documented "Not covered" sustained-outage case.

### Test coverage
Excellent. Regression test reproducing the incident (one 502 mid-poll → exit 0 with the ready branch), the mirror terminal case (404 → terminal after exactly one call), 5xx create-adoption, both repeated for `branch reset`, plus a predicate unit sweep across 500/502/503/504 / 408 / 429 / 4xx / no-status / non-`CLIError`. `CLIError` import present in `create.test.ts:4`. PR reports `npm run lint` green (792 passed, 13 skipped, eslint clean).

### Security & Performance
- **Security:** no new user input reaching SQL/shell/HTTP; no secrets newly logged (debug path still redacts the bearer, `platform.ts:88-95`); no auth/authorization check removed or weakened; no new dependencies (only the 0.2.7 → 0.2.8 version bump).
- **Performance:** retries are bounded by the existing 15-min (branch) / 5-min (deploy) budgets and spaced by `setTimeout(POLL_INTERVAL_MS)`; no busy-loop, no N+1, no new per-request allocation in a hot path.

### Verdict
**approved** (informational — human approval via the separate approve flow). Zero Critical findings; the two notes above are non-blocking. Scope is disciplined — the diff does exactly what the ticket asks and nothing more. Nice work, particularly the care to keep the deploy path bit-for-bit equivalent while extracting the shared predicate.

<sub>🤖 Automated review by the InsForge review bot.</sub>

jwfing (dismissed):
LGTM - approved.

jwfing (commented):
## Review: fix(branch): survive a transient control-plane 502 while polling (v0.2.8)

**Summary:** A well-scoped, well-tested resilience fix that makes the `branch create`/`branch reset` pollers (and the create `POST`) tolerate transient control-plane failures instead of orphaning a real, billing branch behind a non-zero exit — no blocking issues found.

### Requirements context
No `/docs/superpowers/` directory exists in this repo, and the only specs under `docs/specs/` cover the `diagnose` and `db-migrations` commands — **no matching spec/plan found**, so this was assessed against the PR description, the linked agent-e2e incident runs, and the surrounding code.

### Findings

#### Critical
(none)

#### Suggestion
- **Functionality — `src/commands/branch/create.ts:337-353` (timeout fallback can mislabel a just-ready branch).** When the 15-minute budget is exhausted *and* the final re-check itself 502s, the poll returns the stale `lastBranch`. If the branch flipped to `ready` right at the deadline, the command reports it as "still `<non-ready>`" and exits non-zero. This is an extreme double-edge (budget exhaustion + final read transiently failing), it still prints the id/appkey so the caller can recover, and the code comment calls it out honestly ("the worst case is a stale label on a failure"). Not blocking — noting it because the one case it *can* invert is a genuine success, not only "a failure that was already going to fail." A follow-up could re-attempt the final read a few times before falling back.

#### Information
- **Software engineering — test coverage is strong.** The regression test reproduces the reported incident (one 502 mid-poll → still exits 0 with the ready branch), the mirror case (404 stays terminal after exactly one call), 5xx adoption vs. plain-500 non-adoption on create, the predicate unit tests span 5xx/408/429/4xx/no-status/non-`CLIError`, and `reset` gets both the survive-502 and the fail-loudly-when-unconfirmable cases. This is exactly the coverage that would have caught the original bug.
- **Software engineering — `src/lib/errors.ts:72-77` vs `src/commands/branch/create.ts:235-239`.** Two nearby predicates handle 5xx differently on purpose: `isTransientApiError` retries *all* 5xx (including 500) for reads, while `isAmbiguousCreateFailure` deliberately narrows adoption to proxy statuses `{502,503,504}` and excludes 500/408/429. The asymmetry is intentional and thoroughly documented at both sites (a plain 500 is the app's own answer; adopting on it would widen the same-name collision window) — flagging only so a future reader doesn't "unify" them.
- **Performance / behavior change — sustained-outage cost.** Previously a single 5xx aborted the poll immediately; now a *sustained* control-plane 5xx keeps the create poller reading every 3s for the full 15-minute budget (~300 GETs) before timing out. This is the intended tradeoff and the PR documents it under "Not covered." Bounded by `POLL_TIMEOUT_MS`, no unbounded loop.
- **Functionality — verified the "nothing else changes behavior" claim.** `platformFetch` now attaches `statusCode` to the `CLIError` it throws. I checked every `err.statusCode === …` branch in the repo (`config/apply|export|plan`, `diagnose/advisor` `isOssAdvisorRouteMissing`, `domains/index`, `functions/deploy`, `apify-token`) — all of them branch on `ossFetch` errors, and the advisor's `platformFetch` calls use `passThroughStatuses`/`res.status` rather than a thrown-error `statusCode`. The only new readers of a platform-error `statusCode` are command telemetry (benign, arguably more accurate) and the new poll logic. Claim holds.
- **Security.** No security-relevant changes: no new user input reaching SQL/shell/HTTP, no secrets or PII newly logged (error messages carry only the already-known `branchId` and an HTTP status), no auth/authz change, and no new dependencies (version-only bump 0.2.7 → 0.2.8).
- **Software engineering — clean refactors.** `NETWORK_ERROR_CODE` moved to `lib/errors.ts` and re-exported from `lib/api/platform.ts` to avoid a circular import; `deployments/deploy.ts` drops its inline `TRANSIENT_4XX_STATUSES` copy and shares the helper, with a precise comment explaining why the `ossFetch` call site keeps raw fetch rejections retryable (`err instanceof CLIError && !isTransientApiError(err)`), preserving prior behavior for non-`CLIError` throws.

### Verdict
**approved** (informational — the human still gives the explicit GitHub approval via the approve flow). Zero Critical findings; the Suggestion and Information items are non-blocking. Posted as a `COMMENT` per the verdict rule.

jwfing (approved):
LGTM - approved.

